PR #1, opened 2018-10-22
Wavefront Go SDK

vikramraman:


internal/proxy.go
```diff
@@ -0,0 +1,104 @@
...
+	return nil
+}
+
+func (handler *ProxyConnectionHandler) Connected() bool {
+	handler.mtx.RLock()
+	defer handler.mtx.RUnlock()
+	return handler.conn != nil
+}
+
+func (handler *ProxyConnectionHandler) Close() {
```
prydin: ```Close()``` should call ```Flush() ``` or you will lose what's in the buffer.

internal/lines.go
```diff
@@ -0,0 +1,100 @@
...
+	}()
+}
+
+func (lh *LineHandler) HandleLine(line string) error {
+	lh.mtx.Lock()
+	defer lh.mtx.Unlock()
+
+	if len(lh.lines) >= lh.MaxBufferSize {
+		atomic.AddInt64(&lh.failures, 1)
+		return fmt.Errorf("buffer full, dropping line: %s", line)
```
prydin: Wouldn't it be better to try to flush the buffer instead of dropping the line? The whole design of this class would be a lot easier if you used a channel as the underlying buffer. Then ```HandleLine()``` could just send something on the channel and you could have two flavors of ```HandleLine```: One that blocks if the buffer is full and one that drops. Also, this would eliminate the need for the mutex.

internal/lines.go
```diff
@@ -0,0 +1,100 @@
...
+	MaxBufferSize int
+	FlushTicker   *time.Ticker
+	Format        string
+	failures      int64
+	mtx           sync.Mutex
+	lines         []string
+}
+
+func (lh *LineHandler) Start() {
+	go func() {
```
prydin: If it were me, I would let this goroutine take care of the buffering as well as the flushing. You could eliminate the ```lines``` buffer and make it a ```chan(string)``` instead. Create it with your buffer size, and you have a goroutine-safe way of handling all of this with a lot less code.

internal/lines.go
```diff
@@ -0,0 +1,100 @@
...
+		lh.lines = append(lh.lines, drainLines...)
+	}
+}
+
+func (lh *LineHandler) GetFailureCount() int64 {
+	return atomic.LoadInt64(&lh.failures)
+}
+
+func (lh *LineHandler) Stop() {
+	lh.FlushTicker.Stop()
```
prydin: This will leak goroutines! ```Ticker.Stop()``` only stops the clock, meaning that there are no more ticks sent. To assure that the goroutine exits, you have to close the channel. An even better design would be what I'm proposing above, where you'd have a channel that's serving as a buffer. You would close that channel when you're done sending (i.e. in ```Close()```)

internal/proxy.go
```diff
@@ -0,0 +1,104 @@
...
+		return err
+	}
+	return nil
+}
+
+func (handler *ProxyConnectionHandler) resetConnection() {
+	log.Println("resetting wavefront proxy connection")
+	handler.mtx.Lock()
+	defer handler.mtx.Unlock()
+	handler.conn = nil
```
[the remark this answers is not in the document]
vikramraman: The higher level ProxySender calls `isConnected()` and `Connect()` in it's send methods. Setting the connection to nil, ensures isConnected will return false.

marcosrmendezthd:
hi. any idea when we'll have a working version merged? :smile: i'd like to start testing it out.

vikramraman:
@marcosrmendezthd Awesome. We're close. Likely later this week or early next week at the latest.

marcosrmendezthd:
> @marcosrmendezthd Awesome. We're close. Likely later this week or early next week at the latest.

👍 thanks!

marcosrmendezthd:
The other thing I would ask is that if possible is to make the internal package a public one. We will be processing bulk metrics and we've seen that usually the simplest code tends to work best.

vikramraman:
@marcosrmendezthd Thanks for the feedback! Probably not as part of this PR, but definitely open to it. There's nothing preventing you from using it currently, but there are no concrete interfaces backing some of the internal code and it could change. If you find specific use cases that are not covered by the top level interfaces, we can figure out the best way to expose it.

vikramraman:
Thanks for the feedback @prydin . Will look into in a follow up checkin.

PR #2, opened 2018-11-12
Histograms initial commit

laullon:


Gopkg.toml
```diff
@@ -0,0 +1,34 @@
+# Gopkg.toml example
```
vikramraman: Let's remove the commented out example parts.

metrics/histogram.go
```diff
@@ -0,0 +1,112 @@
+package wavefront
```
vikramraman: Put this under `histograms/histogram.go` and change package to histogram.

In terms of terminology histogram is different than metrics. They are two different products that Wavefront now offers.

vikramraman:
Will look into it more later. Just added few minor comments for now.

vikramraman:
Also please add unit tests for your code. It'd be required prior to merging.

metrics/histogram.go
```diff
@@ -0,0 +1,112 @@
...
+
+	tdigest "github.com/caio/go-tdigest"
+)
+
+type Histogram struct {
+	mutex              sync.Mutex
+	priorTimedBinsList []*timedBin
+	currentTimedBin    *timedBin
+
+	metricName string
```
vikramraman: This doesn't belong here. Let's remove the metricName, source and tags from this. This should be purely about creating and updating a histogram. The reporting part such as name etc can be handled in user code.

metrics/histogram.go
```diff
@@ -0,0 +1,112 @@
...
+
+	return h
+}
+
+func (h *Histogram) Add(v float64) {
+	h.rotateCurrentTDigestIfNeedIt()
+	h.currentTimedBin.tdigest.Add(v)
+}
+
+func (h *Histogram) Report(sender senders.Sender) {
```
vikramraman: Remove this method as well.

metrics/histogram.go
```diff
@@ -0,0 +1,112 @@
...
+	timestamp time.Time
+}
+
+type distribution struct {
+	centroids []senders.Centroid
+	timestamp time.Time
+}
+
+func NewHistogram(metricName string, source string, tags map[string]string) *Histogram {
+	h := &Histogram{}
```
vikramraman: For consistency, please use the struct form :

```
&Histogram {
MaxBins: 10,
Granularity: senders.Minute,
Compression: 5,
}
```

metrics/histogram.go
```diff
@@ -0,0 +1,112 @@
...
+func (h *Histogram) Report(sender senders.Sender) {
+	distributions := h.getDistributions()
+	log.Printf("metric: '%s' count:'%d'", h.metricName, len(distributions))
+	hgs := map[senders.HistogramGranularity]bool{h.Granularity: true}
+	for _, distribution := range distributions {
+		sender.SendDistribution(h.metricName, distribution.centroids, hgs, distribution.timestamp.Unix(), h.source, h.tags)
+	}
+}
+
+func (h *Histogram) getDistributions() []distribution {
```
vikramraman: rename to `distributions()`. It's idiomatic in go to omit the get prefixes getters: https://golang.org/doc/effective_go.html#Getters

histograms/histogram.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"sync"
+	"time"
+
+	"github.com/wavefronthq/wavefront-sdk-go/senders"
+
+	tdigest "github.com/caio/go-tdigest"
+)
+
+type Histogram struct {
```
vikramraman: We should define an interface named `Histogram` with the relevant exported methods and have this struct just implement that type. Can rename this struct to something else.

histograms/histogram.go
```diff
@@ -0,0 +1,83 @@
...
+	mutex              sync.Mutex
+	priorTimedBinsList []*timedBin
+	currentTimedBin    *timedBin
+
+	Granularity time.Duration
```
vikramraman: Would be better to keep the struct immutable by not exporting these variables. Let's make them all lower case and just have the `NewHistogram` constructor take these as input and then construct the histogram struct internally and return an interface.

histograms/histogram.go
```diff
@@ -0,0 +1,83 @@
...
+	tdigest   *tdigest.TDigest
+	timestamp time.Time
+}
+
+type Distribution struct {
+	Centroids []senders.Centroid
+	Timestamp time.Time
+}
+
+func NewHistogram() *Histogram {
```
vikramraman: As commented above, take relevant values as input and return the `Histogram` interface.

vikramraman: Refer to https://github.com/rcrowley/go-metrics/blob/master/histogram.go for example.

histograms/histogram.go
```diff
@@ -0,0 +1,83 @@
...
+	h := &Histogram{MaxBins: 10, Granularity: time.Minute, Compression: 5}
+	return h
+}
+
+func (h *Histogram) Add(v float64) {
```
vikramraman: Rename to `Update`.

histograms/histogram.go
```diff
@@ -0,0 +1,83 @@
...
+	MaxBins     int
+}
+
+type timedBin struct {
+	tdigest   *tdigest.TDigest
+	timestamp time.Time
+}
+
+type Distribution struct {
+	Centroids []senders.Centroid
```
vikramraman: Let's move all histogram related types into this `histogram` package from the `senders` package. Put them under a `types.go` file.

vikramraman:
Overall the histo implementation LGTM. We need to expose methods to get the histogram values though:
`Min()`, `Max()`, `Mean()`, `Count()`, `Sum()`, `Quantile(double quantile)` etc. And also `Snapshot()` which will return a readonly copy of the histogram.

laullon:
We already have a `Snapshot()`, it's called `Distributions()` it give you all the info on the T-Digest, is not read-only, but the modifications will not affect to the histogram.

histogram/histogram.go
```diff
@@ -0,0 +1,219 @@
...
+
+	max := math.SmallestNonzeroFloat64
+	h.currentTimedBin.tdigest.ForEachCentroid(func(mean float64, count uint32) bool {
+		max = math.Max(max, mean)
+		return true
+	})
+	return max
+}
+
+// Max returns the minimun Value of samples on this histogram.
```
vikramraman: typo: Min

histogram/histogram.go
```diff
@@ -0,0 +1,219 @@
...
+
+	min := math.MaxFloat64
+	h.currentTimedBin.tdigest.ForEachCentroid(func(mean float64, count uint32) bool {
+		min = math.Min(min, mean)
+		return true
+	})
+	return min
+}
+
+// Max returns the sum of all values on this histogram.
```
vikramraman: typo: Sum

vikramraman:
@laullon Just some minor typo's in comments. Please fix them.

it'd also be good to add some unit tests for the min, max, mean etc. But I'm okay with adding those in a following checkin.

marcosrmendezthd:
hi! are we doing a release with this new functionality? using go.mod i'm stuck on v0.9.1 😄

PR #10, opened 2019-01-28
exposing Histogram Granularity and Heartbeater

laullon:


laullon:
need it by https://github.com/wavefrontHQ/go-metrics-wavefront/pull/12

histogram/types.go
```diff
@@ -10,19 +12,32 @@ type Centroid struct {
 type HistogramGranularity int8
 
 const (
-	MINUTE HistogramGranularity = iota
+	SECOND HistogramGranularity = iota
```
vikramraman: Remove this? Shouldn't be exporting this externally as we only support minute, hour and day bins.

vikramraman:
minor comment. lgtm otherwise.

heartbeater/heartbeater.go
```diff
@@ -0,0 +1,64 @@
+package heartbeater
```
vikramraman: Let's put this under the application package. Since it's used to detect applications.

heartbeater/heartbeater.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func (hb *heartbeater) Close() {
+	hb.stop <- true
+}
+
+func (hb *heartbeater) beat() {
+	tags := hb.application.Map()
+	tags["component"] = "wavefront-generated"
+	error := hb.sender.SendMetric("~component.heartbeat", 1, 0, hb.source, tags)
```
vikramraman: rename to `err` for consistency.

heartbeater/heartbeater.go
```diff
@@ -0,0 +1,64 @@
...
+	if error != nil {
+		log.Printf("heartbeater SendMetric error: %v\n", error)
+	} else {
+		log.Printf("heartbeater beat OK\n")
```
vikramraman: Can remove the else part. In general we want to minimize logging unless it's an error.

histogram/types.go
vikramraman: @laullon Will you be removing this?

laullon: I use it for unit testing... will try to find another way, and then remove it

vikramraman:
Lgtm. Thanks!

PR #16, opened 2019-03-12
Fix concurrency issues and other optimizations

mshvol:
**Fixes & Changes**

- [x] Connected-Connect race in Proxy Sender
- [x] LineHandler Flush race/panic due to bufLen
- [x] Flush in right order: ProxyConnHandler, LineHandler
- [x] FlushAll added to LineHandler to flush all lines on Close()
- [x] resetConnection close (Finalizer will close it but Conn lingers until GC'd)
- [x] HistoLine preprocessing
- [x] Heartbeater Ticker leak
- [x] Unnecessary NewBufferString 
- [x] Signal channels: ProxyConnHandler, LineHandler, Heartbeater
- [x]  Ticker refactoring in handlers
- [x]  Readme edits
- [x]  Initial testing framework for senders (partially automated)

### DIFF
--- a/application/heartbeater.go
+++ b/application/heartbeater.go
@@ -19,7 +19,7 @@ type heartbeater struct {
 	components  []string
 
 	ticker *time.Ticker
-	stop   chan bool
+	stop   chan struct{}
 }
 
 // StartHeartbeatService will create and start a new HeartbeatService
@@ -30,7 +30,7 @@ func StartHeartbeatService(sender senders.Sender, application Tags, source strin
 		source:      source,
 		components:  components,
 		ticker:      time.NewTicker(5 * time.Minute),
-		stop:        make(chan bool, 1),
+		stop:        make(chan struct{}),
 	}
 
 	go func() {
@@ -49,7 +49,8 @@ func StartHeartbeatService(sender senders.Sender, application Tags, source strin
 }
 
 func (hb *heartbeater) Close() {
-	hb.stop <- true
+	hb.ticker.Stop()
+	hb.stop <- struct{}{} // block until goroutine exits
 }
 
 func (hb *heartbeater) beat() {
--- a/internal/lines.go
+++ b/internal/lines.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"time"
 )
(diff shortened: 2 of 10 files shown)


application/heartbeater.go
```diff
@@ -30,7 +30,7 @@ func StartHeartbeatService(sender senders.Sender, application Tags, source strin
 		source:      source,
 		components:  components,
 		ticker:      time.NewTicker(5 * time.Minute),
-		stop:        make(chan bool, 1),
+		stop:        make(chan struct{}),
```
laullon: why?

mshvol: `struct{}` is the smallest data type available in go with 0 bytes.
https://dave.cheney.net/2014/03/25/the-empty-struct

mshvol: If you're asking why unbuffered, Hearbeater.Close() will block until goroutine exits.

laullon: so, your are saving 1 byte of memory, and IMO making the code harder to read

prydin:
chan struct{} is very common in Go. It's the recommended way to declare a channel carrying no data (i.e. just used for event signalling)

application/heartbeater.go
prydin: chan struct{} is very common in Go. It's the recommended way to declare a channel carrying no data (i.e. just used for event signalling)

laullon: Ok, clear.

internal/lines.go
```diff
@@ -61,6 +75,20 @@ func (lh *LineHandler) Flush() error {
...
+func (lh *LineHandler) FlushAll() error {
+	lh.mtx.Lock()
+	defer lh.mtx.Unlock()
+	bufLen := len(lh.buffer)
+	if bufLen > 0 {
+		lines := make([]string, bufLen)
+		for i := 0; i < bufLen; i++ {
+			lines[i] = <-lh.buffer
+		}
+		return lh.report(lines)
```
vikramraman: We should limit to the batch size even when we flush the entire buffer. We can delegate to Flush() until the buffer is drained. But bail out if there is an error when reporting to avoid an infinite loop as points are added back to buffer in that case.

senders/proxy.go
```diff
@@ -73,11 +73,9 @@ func (sender *proxySender) SendMetric(name string, value float64, ts int64, sour
 		return fmt.Errorf("proxy metrics port not provided, cannot send metric data")
 	}
 
-	if !sender.metricHandler.Connected() {
-		err := sender.metricHandler.Connect()
-		if err != nil {
-			return err
-		}
+	err := sender.metricHandler.Connect()
```
vikramraman: `Connected()` uses a read lock, whereas `Connect()` uses a write lock. It'd better to retain the previous approach for concurrent operations.

internal/lines.go
mshvol: Actually, now that I think about it, why should Flush() even flush partially? Isn't it better if it flushes everything in batches until buffer is empty?

PR #31, opened 2019-09-05
support sending events using direct ingestion and proxy

laullon:


senders/direct.go
```diff
@@ -55,6 +62,7 @@ func NewDirectSender(cfg *DirectConfiguration) (Sender, error) {
 	sender.histoHandler = makeLineHandler(reporter, cfg, histogramFormat, "histograms", sender.internalRegistry)
 	sender.spanHandler = makeLineHandler(reporter, cfg, traceFormat, "spans", sender.internalRegistry)
 	sender.spanLogHandler = makeLineHandler(reporter, cfg, spanLogsFormat, "span_logs", sender.internalRegistry)
+	sender.eventHandler = makeLineHandler(eventReporter, cfg, "event", "events", sender.internalRegistry)
```
vikramraman: declare a constant for "event" format.

internal/direct.go
```diff
@@ -21,50 +22,96 @@ const (
 	authzHeader     = "Authorization"
 	bearer          = "Bearer "
 	gzipFormat      = "gzip"
-	octetStream     = "application/octet-stream"
```
vikramraman: Let's still retain these constants and use them as needed rather than use string literals in the constructor.

senders/direct.go
```diff
@@ -63,6 +71,10 @@ func NewDirectSender(cfg *DirectConfiguration) (Sender, error) {
 func makeLineHandler(reporter internal.Reporter, cfg *DirectConfiguration, format, prefix string,
 	registry *internal.MetricRegistry) *internal.LineHandler {
 	flushInterval := time.Second * time.Duration(cfg.FlushIntervalSeconds)
+	if format == "event" {
+		return internal.NewLineHandler(reporter, format, flushInterval, 1, cfg.MaxBufferSize,
```
vikramraman: If the only thing that differs is the batch size, maybe for clarity just do:
```
batchSize := cfg.BatchSize
if format == event {
    batchSize = 1
}
return internal.NewLineHandler(...)
```

internal/direct.go
```diff
@@ -21,50 +22,96 @@ const (
...
+		contentType: "application/octet-stream",
+		gzip:        true,
+	}
+	for _, setter := range setters {
+		setter(dr)
+	}
+	return dr
 }
 
 func (reporter directReporter) Report(format string, pointLines string) (*http.Response, error) {
```
vikramraman: It'd be cleaner to have a separate method for reporting events and invoke that from the LineHandler based on the format. The `Report` method was really meant to handle the `/report?f=` endpoint. Let's leave that as is and add a new one for reporting events.

internal/direct.go
```diff
@@ -55,17 +62,50 @@ func (reporter directReporter) Report(format string, pointLines string) (*http.R
...
+	serverURL string
+	token     string
+}
+
+// NewDirectEventReporter create a event Reporter
+func NewDirectEventReporter(server string, token string) Reporter {
+	return &directEventReporter{serverURL: server, token: token}
+}
+
+func (reporter directEventReporter) Report(format string, event string) (*http.Response, error) {
```
vikramraman: Could this just be moved as `ReportEvent` within the `directReporter` struct?

laullon: ok

senders/direct.go
```diff
@@ -123,12 +136,25 @@ func (sender *directSender) SendSpan(name string, startMillis, durationMillis in
 	return nil
 }
 
+func (sender *directSender) SendEvent(name string, startMillis, endMillis int64, source string, tags map[string]string, setters ...event.Option) error {
+	line, err := EventLineJSOM(name, startMillis, endMillis, source, tags, setters...)
+	if err != nil {
+		return err
+	}
+	err = sender.eventHandler.HandleLine(line)
```
vikramraman: simply `return sender.eventHandler.HandleLin(line)`

internal/direct.go
vikramraman: @laullon Previous comment is still not addressed. Let's remove the `directEventReporter` struct and just add a `ReportEvent` method to the existing `Reporter` interface/implementation. 

Basically the `Reporter` interface is a thing wrapper around the WF API and will support the `Report` (/report) and `ReportEvent` (/api/v2/event) endpoints.

internal/interfaces.go
```diff
@@ -7,7 +7,6 @@ import "net/http"
 // Reporter is an interface for reporting data to a Wavefront service.
 type Reporter interface {
 	Report(format string, pointLines string) (*http.Response, error)
```
vikramraman: Add a `ReportEvent` method to this interface

internal/lines.go
```diff
@@ -72,6 +77,15 @@ func (lh *LineHandler) Start() {
 				err := lh.Flush()
 				if err != nil {
 					log.Println(err)
+					if err == errThrottled {
```
vikramraman: Hmmm, this would impact the other atoms as well (metrics, spans etc) as we are only looking for a 406 status code. We should really try to isolate this to just the event handling.

I would suggest creating a separate ErrorHandler interface that is an optional setter for LineHandler (only leveraged when creating the eventHandler). And LineHandler delegates to this here and in the report method if it is non-nil.

senders/proxy.go
```diff
@@ -147,6 +148,25 @@ func (sender *proxySender) SendSpan(name string, startMillis, durationMillis int
 	return nil
 }
 
+func (sender *proxySender) SendEvent(name string, startMillis, endMillis int64, source string, tags map[string]string, setters ...event.Option) error {
+	if sender.metricHandler == nil {
```
vikramraman: Create a separate eventHandler and use that. Events will be on a different port than the metrics. Should also change the `ProxyConfiguration` to support an `EventsPort`.

internal/lines.go
laullon: this is just to show a log message

senders/proxy.go
laullon: right now is the same port

internal/direct.go
laullon: this change will add a lot of if's that we removed on the previous comment...

senders/proxy.go
```diff
@@ -114,17 +121,18 @@ func (sender *proxySender) SendDistribution(name string, centroids []histogram.C
...
 		return err
 	}
-	err = sender.histoHandler.SendData(line)
+	err = handler.SendData(line)
 	return err
 }
 
 func (sender *proxySender) SendSpan(name string, startMillis, durationMillis int64, source, traceId, spanId string, parents, followsFrom []string, tags []SpanTag, spanLogs []SpanLog) error {
-	if sender.spanHandler == nil {
+	handler := sender.handlers[histoHandler]
```
vikramraman: This needs to be `spanHandler`

senders/proxy.go
```diff
@@ -143,18 +151,19 @@ func (sender *proxySender) SendSpan(name string, startMillis, durationMillis int
...
-		return sender.spanHandler.SendData(logs)
+		return handler.SendData(logs)
 	}
 	return nil
 }
 
 func (sender *proxySender) SendEvent(name string, startMillis, endMillis int64, source string, tags map[string]string, setters ...event.Option) error {
-	if sender.metricHandler == nil {
-		return errors.New("proxy metrics port not provided, cannot send metric data")
+	handler := sender.handlers[histoHandler]
```
vikramraman: This needs to be `eventHandler`

PR #32, opened 2019-09-25
support customTags for heartbeat

djia-vm-wf:
#28

### DIFF
--- a/application/heartbeater.go
+++ b/application/heartbeater.go
@@ -86,15 +86,11 @@ func (hb *heartbeater) send(tags map[string]string) {
 
 func (hb *heartbeater) AddCustomTags(tags map[string]string) {
 	hb.mux.Lock()
-	exist := false
+	defer hb.mux.Unlock()
 	for _, existCustomTag := range hb.customTags {
 		if reflect.DeepEqual(existCustomTag, tags) {
-			exist = true
-			break
+			return
 		}
 	}
-	if !exist {
-		hb.customTags = append(hb.customTags, tags)
-	}
-	hb.mux.Unlock()
+	hb.customTags = append(hb.customTags, tags)
 }
--- a/application/heartbeater.go
+++ b/application/heartbeater.go
@@ -10,6 +10,7 @@ import (
 // HeartbeatService sends a heartbeat metric every 5 minutes
 type HeartbeatService interface {
 	Close()
+	Send(tags map[string]string)
 }
 
 type heartbeater struct {
@@ -56,15 +57,15 @@ func (hb *heartbeater) Close() {
 func (hb *heartbeater) beat() {
 	tags := hb.application.Map()
 	tags["component"] = "wavefront-generated"
-	hb.send(tags)
+	hb.Send(tags)
 
 	for _, component := range hb.components {
 		tags["component"] = component
-		hb.send(tags)
+		hb.Send(tags)
 	}
 }
 
-func (hb *heartbeater) send(tags map[string]string) {
+func (hb *heartbeater) Send(tags map[string]string) {
 	err := hb.sender.SendMetric("~component.heartbeat", 1, 0, hb.source, tags)
 	if err != nil {
 		log.Printf("heartbeater SendMetric error: %v\n", err)
--- a/application/heartbeater.go
+++ b/application/heartbeater.go
@@ -2,6 +2,7 @@ package application
 
 import (
 	"log"
+	"reflect"
 	"sync"
 	"time"
 
@@ -19,8 +20,8 @@ type heartbeater struct {
 	application Tags
 	source      string
 	components  []string
-	customTags  []map[string]string
 	mux         sync.Mutex
+	customTags  []map[string]string
 
 	ticker *time.Ticker
 	stop   chan struct{}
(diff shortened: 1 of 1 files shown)


application/heartbeater.go
```diff
@@ -56,11 +57,11 @@ func (hb *heartbeater) Close() {
 func (hb *heartbeater) beat() {
 	tags := hb.application.Map()
 	tags["component"] = "wavefront-generated"
-	hb.send(tags)
+	hb.Send(tags)
```
vikramraman: Revert this? The internal function is lowercase `send`. Why is this capitalized?

application/heartbeater.go
```diff
@@ -56,11 +57,11 @@ func (hb *heartbeater) Close() {
...
 
 	for _, component := range hb.components {
 		tags["component"] = component
-		hb.send(tags)
+		hb.Send(tags)
```
vikramraman: Ditto. Revert?

vikramraman:
I think custom tags are already supported in the heartbeater via this line: 
https://github.com/wavefrontHQ/wavefront-sdk-go/blob/444c396a44fc920294d5e7f960f6a8fc7d8a5ad2/application/heartbeater.go#L57.

Don't think we need to add a separate `AddCustomTags` method?

application/heartbeater.go
```diff
@@ -70,3 +71,9 @@ func (hb *heartbeater) send(tags map[string]string) {
 		log.Printf("heartbeater SendMetric error: %v\n", err)
 	}
 }
+
+func (hb *heartbeater) AddCustomTags(tags map[string]string) {
+	for key, _ := range tags {
+		hb.application.CustomTags[key] = tags[key]
```
vikramraman: let's maintain a separate set of maps and send the heartbeat metrics for each map.

application/heartbeater.go
vikramraman: before we send app tags, add a separate logic to send the custom tags.

application/heartbeater.go
```diff
@@ -70,3 +71,9 @@ func (hb *heartbeater) send(tags map[string]string) {
 		log.Printf("heartbeater SendMetric error: %v\n", err)
 	}
 }
+
+func (hb *heartbeater) AddCustomTags(tags map[string]string) {
+	for key, _ := range tags {
```
vikramraman: need to add a mutex around this and during the send.

application/heartbeater.go
```diff
@@ -10,13 +11,16 @@ import (
...
+	AddCustomTags(tags map[string]string)
 }
 
 type heartbeater struct {
 	sender      senders.Sender
 	application Tags
 	source      string
 	components  []string
+	customTags  []map[string]string
+	mux         sync.Mutex
```
vikramraman: move this above the customTags. It's idiomatic in Go to define the mutex above the fields it provides mutually exclusive access to.

application/heartbeater.go
```diff
@@ -70,3 +82,9 @@ func (hb *heartbeater) send(tags map[string]string) {
 		log.Printf("heartbeater SendMetric error: %v\n", err)
 	}
 }
+
+func (hb *heartbeater) AddCustomTags(tags map[string]string) {
+	hb.mux.Lock()
+	hb.customTags = append(hb.customTags, tags)
```
vikramraman: Since we do not have a Set, and are using a slice instead, we need to verify we are not appending the same map[k]->values that already exist in our list of customTags.

We should not simply append the tags. Only append if the input tags map is different than what we already have amongst `hb.customTags`.

application/heartbeater.go
```diff
@@ -62,6 +65,15 @@ func (hb *heartbeater) beat() {
 		tags["component"] = component
 		hb.send(tags)
 	}
+
+	//send customTags heartbeat
+	hb.mux.Lock()
+	for len(hb.customTags) > 0 {
+		tags := hb.customTags[0]
+		hb.customTags = hb.customTags[1:]
```
vikramraman: This mutates the internal customTags. It will be empty the next time `beat()` runs.

djia-vm-wf: it should be empty

application/heartbeater.go
```diff
@@ -70,3 +83,18 @@ func (hb *heartbeater) send(tags map[string]string) {
...
+	for _, existCustomTag := range hb.customTags {
+		if reflect.DeepEqual(existCustomTag, tags) {
+			exist = true
+			break
+		}
+	}
+	if !exist {
+		hb.customTags = append(hb.customTags, tags)
+	}
+	hb.mux.Unlock()
```
vikramraman: Can move this just after the lock statement and have `defer hb.mux.Unlock()`.

application/heartbeater.go
```diff
@@ -70,3 +83,18 @@ func (hb *heartbeater) send(tags map[string]string) {
 		log.Printf("heartbeater SendMetric error: %v\n", err)
 	}
 }
+
+func (hb *heartbeater) AddCustomTags(tags map[string]string) {
+	hb.mux.Lock()
+	exist := false
+	for _, existCustomTag := range hb.customTags {
+		if reflect.DeepEqual(existCustomTag, tags) {
+			exist = true
```
vikramraman: Can simply return if it already exists. No need for the `exists` boolean.

PR #42, opened 2019-12-17
Djia/sanitize

djia-vm-wf:


senders/formatter.go
```diff
@@ -308,3 +314,52 @@ func isUUIDFormat(str string) bool {
 	}
 	return true
 }
+
+func sanitizeInternal(str string) string {
```
vikramraman: Could you add some comments to this method and `sanitizeValue` describing what it does? It'd be useful to someone looking at this independently to understand what we're doing.

senders/formatter.go
```diff
@@ -308,3 +314,52 @@ func isUUIDFormat(str string) bool {
...
+			sb.WriteString("-")
+		}
+	}
+	return sb.String()
+}
+
+func sanitizeValue(str string) string {
+	res := strings.TrimSpace(str)
+	if strings.Contains(str, "\"") || strings.Contains(str, "'") {
+		resReg, _ := regexp.Compile("\"")
```
vikramraman: Make these global vars and pre-compile the regular expression. You can then reuse it within this method. Compiling it for every call is expensive and should be avoided.

Use `MustCompile(...)` for initializing the global vars.

senders/formatter.go
```diff
@@ -308,3 +314,52 @@ func isUUIDFormat(str string) bool {
...
+		res = resReg.ReplaceAllString(res, "\\\"")
+	}
+	resReg, _ := regexp.Compile("\\n")
```
vikramraman: Ditto. Make it a global var.

senders/formatter.go
djia-vm-wf: these two can actually be const. made them const

PR #43, opened 2020-01-14
Added check to ignore delta counter with value 0

srinivas-kandula:
As 0 is not a valid value for delta counter added a check to ignore this.

### DIFF
--- a/senders/direct.go
+++ b/senders/direct.go
@@ -95,7 +95,10 @@ func (sender *directSender) SendDeltaCounter(name string, value float64, source 
 	if !internal.HasDeltaPrefix(name) {
 		name = internal.DeltaCounterName(name)
 	}
-	return sender.SendMetric(name, value, 0, source, tags)
+	if value > 0 {
+	        return sender.SendMetric(name, value, 0, source, tags)
+        }
+	return nil
 }
 
 func (sender *directSender) SendDistribution(name string, centroids []histogram.Centroid,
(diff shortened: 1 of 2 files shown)


senders/direct.go
```diff
@@ -95,7 +95,10 @@ func (sender *directSender) SendDeltaCounter(name string, value float64, source
 	if !internal.HasDeltaPrefix(name) {
 		name = internal.DeltaCounterName(name)
 	}
-	return sender.SendMetric(name, value, 0, source, tags)
+	if value > 0 {
+	        return sender.SendMetric(name, value, 0, source, tags)
+        }
+	return nil
```
vikramraman: We should probably return an error if value == 0.

senders/proxy.go
```diff
@@ -102,7 +102,10 @@ func (sender *proxySender) SendDeltaCounter(name string, value float64, source s
 	if !internal.HasDeltaPrefix(name) {
 		name = internal.DeltaCounterName(name)
 	}
-	return sender.SendMetric(name, value, 0, source, tags)
+	if value > 0 {
+	        return sender.SendMetric(name, value, 0, source, tags)
+	}
+	return nil
```
vikramraman: Ditto. Return error if value == 0

vikramraman:
@srinivas-kandula ignore my previous comments. Current implementation is fine. Let's fix the formatting issue and we can merge this PR.

PR #44, opened 2020-01-29
Update README.md

howardyoo:
Added additional instructions that you need to have go.mod file in order to successfully compile and run wavefront go sdk.


vikramraman:
@howardyoo Thanks for the PR. But we don't want to specify a particular dependency management requirement in the README. Users are free to use any of go mod, dep etc.

PR #47, opened 2020-05-21
Enhance the wavefront-sdk-go to support Kubernetes environments

laullon:


application/application.go
```diff
@@ -23,15 +30,44 @@ func New(application, service string) Tags {
 }
 
 // Map with all values
-func (a Tags) Map() map[string]string {
+func (app *Tags) Map() map[string]string {
+	os.Getenv("pp")
```
vikramraman, line 34: is this necessary? remove?

laullon: oh, sorry...  ;-)

application/application.go
```diff
@@ -23,15 +30,44 @@ func New(application, service string) Tags {
...
 		allTags[k] = v
 	}
 	return allTags
 }
+
+// AddCustomTagsFromEnv set additional custom tags from environment variables that match the given regex.
+func (app *Tags) AddCustomTagsFromEnv(regx string) {
+	r, err := regexp.Compile(regx)
+	if err != nil {
+		log.Printf("Error creating custom tags: %v\n", err)
```
vikramraman, line 50: Return an error here rather than just log it?

application/application.go
```diff
@@ -23,15 +30,43 @@ func New(application, service string) Tags {
...
+			if len(v) > 0 {
+				app.CustomTags[k] = v
+			}
+		}
+	}
+}
+
+// AddCustomTagFromEnv Set a custom tag from the given environment variable.
+func (app *Tags) AddCustomTagFromEnv(varName, tag string) {
+	v := os.Getenv(varName)
```
vikramraman, line 68: Maybe return an error if the env var doesn't exist?

Issue #46, opened 2020-03-26
Deprecate WavefrontProxyClient


vikramraman:
See https://github.com/wavefrontHQ/wavefront-sdk-java/issues/124 for details.

vikramraman:
Addressed by #50

Issue #7, opened 2018-12-07
make formatting functions public


marcosrmendezthd:
If we can promote functions in formatter.go to be public, we can allow users to manage their own connectivity.

vikramraman:
done.

PR #50, opened 2020-05-26
Deprecate WavefrontProxyClient [WIP]

laullon:
Addresses Issue #46


vikramraman:
@laullon At a high level, here's what we need to do:
- Add a construct similar to [WavefrontClientFactory](https://github.com/wavefrontHQ/wavefront-sdk-java/blob/master/src/main/java/com/wavefront/sdk/common/clients/WavefrontClientFactory.java)
- Add a [WavefrontMultiClient](https://github.com/wavefrontHQ/wavefront-sdk-java/blob/master/src/main/java/com/wavefront/sdk/common/clients/WavefrontMultiClient.java)
- Latest versions of the proxy support receiving all atoms (metrics, histograms, spans etc) using a single port over HTTP. So we will deprecate the existing Proxy and Direct senders in favor of the new `WavefrontClientFactory`.

Since both the proxy and direct ingestion will be over HTTP going forward, the configuration for both is similar barring the URL endpoint (batchsize, queuesize, flushinterval etc apply to both).  And it will be backed by a singular [WavefrontClient](https://github.com/wavefrontHQ/wavefront-sdk-java/blob/master/src/main/java/com/wavefront/sdk/common/clients/WavefrontClient.java)

vikramraman:
A lot of the changes break backwards compatibility. It'd be nice to retain backwards compatibility if possible.

internal/formatter.go
```diff
@@ -1,4 +1,4 @@
-package senders
+package internal
```
vikramraman, line 1: There was a request to make these public earlier: https://github.com/wavefrontHQ/wavefront-sdk-go/issues/7

Moving it to a different package breaks backwards compatibility. We should consider deprecating existing code and leveraging Go [aliases](https://medium.com/a-journey-with-go/go-aliases-simple-and-efficient-8506d93b079e) where possible to smoothen the transition.

senders/interfaces.go
```diff
@@ -3,53 +3,10 @@
...
 
 import (
-	"github.com/wavefronthq/wavefront-sdk-go/event"
-	"github.com/wavefronthq/wavefront-sdk-go/histogram"
-	"github.com/wavefronthq/wavefront-sdk-go/internal"
+	"github.com/wavefronthq/wavefront-sdk-go/wavefront"
 )
 
-// MetricSender Interface for sending metrics to Wavefront
-type MetricSender interface {
```
vikramraman, line 12: Again, breaks backwards compatibility. Let's deprecate and use aliases if needed.

wavefront/client.go
```diff
@@ -0,0 +1,209 @@
...
+	"time"
+
+	"github.com/wavefronthq/wavefront-sdk-go/event"
+	"github.com/wavefronthq/wavefront-sdk-go/histogram"
+	"github.com/wavefronthq/wavefront-sdk-go/internal"
+	"github.com/wavefronthq/wavefront-sdk-go/types"
+)
+
+// Client Interface for sending metrics, distributions and spans to Wavefront
+type Client interface {
```
vikramraman, line 16: Let's not introduce a new `Client` interface. `WavefrontSender` will still be the main interface.

`Client` will be the singular implementation (struct) of `WavefrontSender` (instead of the previous direct vs proxy clients). And MultiClient likewise a wrapper around clients that implements the WavefrontSender interface.

internal/formatter.go
```diff
@@ -136,23 +132,23 @@ func SpanLine(name string, startMillis, durationMillis int64, source, traceId, s
...
 
 	sb.WriteString(sanitizeValue(name))
 	sb.WriteString(" source=")
-	sb.WriteString(sanitizeValue(source))
-	sb.WriteString(" traceId=")
-	sb.WriteString(traceId)
-	sb.WriteString(" spanId=")
-	sb.WriteString(spanId)
+	sb.WriteString(strconv.Quote(sanitizeInternal(source)))
+	sb.WriteString(" traceID=")
```
vikramraman, line 148: The line format at the proxy looks for `traceId` and `spanId`. Let's revert these changes.

wavefront/client.go
```diff
@@ -0,0 +1,209 @@
+package wavefront
```
vikramraman, line 1: I don't think we need a new package. We can just add it under `senders`.

Going forward it'll be `senders.NewClient(...)` instead of direct or proxy sender.

senders/client_factory.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if !strings.HasPrefix(strings.ToLower(u.Scheme), "http") {
+		return nil, fmt.Errorf("invalid schema '%s', only 'http' is supported", u.Scheme)
+	}
+
+	if len(u.User.String()) > 0 {
+		cfg.Token = u.User.String()
```
vikramraman, line 49: To maintain consistency with other SDKs (Java, Python etc), let's not encode the token in the URL.

Java provides a couple of [builder methods](https://github.com/wavefrontHQ/wavefront-sdk-java/blob/0f88426e1f827989b79f85d739d9c08c8e7eb58b/src/main/java/com/wavefront/sdk/common/clients/WavefrontClient.java#L133). We can probably just have the constructor take a `server` and `token` strings as input. And add clear comments describing what the server needs to be for proxy vs direct.

senders/proxy.go
```diff
@@ -175,6 +176,7 @@ func (sender *proxySender) SendEvent(name string, startMillis, endMillis int64,
 	if err != nil {
 		return err
 	}
+	println("[proxy]", line)
```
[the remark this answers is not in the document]
laullon: sure

senders/types.go
```diff
@@ -11,7 +16,43 @@ type SpanLog struct {
 }
 
 type SpanLogs struct {
-	TraceId string    `json:"traceId"`
-	SpanId  string    `json:"spanId"`
+	TraceID string    `json:"traceId"`
```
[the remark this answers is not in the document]
laullon: Yeah... the IDE keeps telling me to change it to ID, and I must apply auto change on some point... I already have it reverted ;-)

senders/proxy.go
```diff
@@ -212,3 +214,8 @@ func (sender *proxySender) GetFailureCount() int64 {
 	}
 	return failures
 }
+
+//TODO: remove, just for tesing
```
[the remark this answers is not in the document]
laullon: sure

internal/reporter.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	apiURL := reporter.serverURL + eventEndpoint
+	req, err := http.NewRequest("POST", apiURL, strings.NewReader(event))
+	if err != nil {
+		return &http.Response{}, err
+	}
+
+	req.Header.Set(contentType, applicationJSON)
+	// req.Header.Set(contentEncoding, gzipFormat)
```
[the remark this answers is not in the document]
laullon: There is a problem with the proxy, will check.

senders/client_factory.go
laullon: Java also support it on the URL 
`// Create a factory and add a client with the following URL format: "http://TOKEN@DOMAIN.wavefront.com"`

I can add the token as an `Option`, OK ?

randallt:
How can I use the new Sender that only takes a URL? Looks like
```
import (
	wfsender "github.com/wavefronthq/wavefront-sdk-go/senders"
)

func InitTracing(serviceName string, applicationName string) wfsender.Sender {
	sender, err := wfsender.NewSender("http://wfproxy.wavefront:2878")
	if err != nil {
		panic(err)
	}
	return sender
}
```

currently doesn't work, as NewSender is undefined.

laullon:
This is not yet released, if you want to test it make sure you point your dependencies to master branch.

randallt:
@laullon is there a timeline to release this?

PR #57, opened 2020-07-15
Add build version to sdk internal metrics.

akodali18:


internal/utils.go
```diff
@@ -16,3 +23,29 @@ func min(x, y int) int {
 	}
 	return y
 }
+
+func GetSemVer(version string) (float64, error) {
+	if len(version) > 0 {
```
laullon, line 28: You can simplify this code with:
```
	res := semVerRegex.FindStringSubmatch(version)
	v := fmt.Sprintf("%s.%02s%03s\n", res[1], res[2], res[3])
```

play: https://play.golang.org/p/s-WXMWlfr1I

internal/utils.go
```diff
@@ -1,6 +1,13 @@
...
 
-import "os"
+import (
+	"os"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+var semVerRegex, _ = regexp.Compile("([0-9]\\d*)\\.(\\d+)\\.(\\d+)(?:-([a-zA-Z0-9]+))?")
```
laullon, line 10: Use `regexp.MustCompile`

internal/utils_test.go
```diff
@@ -0,0 +1,43 @@
+package internal
+
+import (
+	"testing"
+)
+
+func TestGetSemVer(t *testing.T) {
+	if sdkVersion, e := GetSemVer(""); sdkVersion != 0 || e != nil {
+		t.Error("Expected sdk version to be 0 but got : ", sdkVersion)
+	}
```
laullon, line 10: Use `assert.Equal` and `assert.Nil` the code will be more clear.

version/version.go
```diff
@@ -0,0 +1,5 @@
+package version
+
+var (
+	Version = "0.9.8"
```
laullon, line 4: Why don't you just hard code the `sdkVersion` value here? that way you can remove all the translation code.

senders/direct.go
```diff
@@ -47,6 +48,13 @@ func NewDirectSender(cfg *DirectConfiguration) (Sender, error) {
 		internal.SetPrefix("~sdk.go.core.sender.direct"),
 		internal.SetTag("pid", strconv.Itoa(os.Getpid())),
```
laullon, line 49: Even more... we can put the version as tag here, like the pid, that way we don't need to report a new metric.

version/version.go
```diff
@@ -0,0 +1,5 @@
...
+)
```
laullon, line 5: Make it const.

senders/direct.go
akodali18: We thought about emitting this as a tag and it was decided that a metric would be more useful going forward like it can be used to create a chart on dashboard etc.

version/version.go
akodali18: This is to keep the code more readable and maintainable by future developers. The sdkVersion reported to wavefront in its gauge format might be confusing to some.

internal/utils.go
akodali18: Thanks, this is a lot simpler, made the change.

internal/utils_test.go
akodali18: Thanks for the suggestion, done.

internal/utils.go
```diff
@@ -16,3 +23,12 @@ func min(x, y int) int {
 	}
 	return y
 }
+
+func GetSemVer(version string) (float64, error) {
+	if len(version) > 0 {
+		res := semVerRegex.FindStringSubmatch(version)
+		sdkVersion := fmt.Sprintf("%s.%02s%02s", res[1], res[2], res[3])
```
vikramraman, line 30: Can we add a length check. `len(res) == 3` for example around this.

vikramraman (approved):

PR #58, opened 2020-07-29
Recompress histograms [WIP]

laullon:


histogram/types.go
```diff
@@ -1,13 +1,39 @@
...
+	tmp := make(map[float64]int)
+	for _, c := range centroids {
+		if _, ok := tmp[c.Value]; ok {
+			tmp[c.Value] += c.Count
+		} else {
+			tmp[c.Value] = c.Count
+		}
+	}
+	for v, c := range tmp {
+		res = append(res, Centroid{Value: v, Count: c})
```
vikramraman, line 31: pre-allocate res to be len(tmp) and avoid appending?

Issue #35, opened 2019-10-21
Add flag to disable SDK internal metrics


haosong:
Similar to the issue for python: https://github.com/wavefrontHQ/wavefront-sdk-python/issues/51

PR #61, opened 2020-11-22
Send Internal Diagnostic Delta Counters

joannak-vmware:
* Update Sanitize to Include Delta Tilda
    * Small README fix for Send Spans

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,9 +224,9 @@ Wavefront supports different metric types, such as gauges, counters, delta count
 sender.SendMetric("new-york.power.usage", 42422.0, 0, "go_test", map[string]string{"env" : "test"})
 
 // Wavefront delta counter format
-// <metricName> <metricValue>  [<timestamp>] source=<source> [pointTags]
+// <metricName> <metricValue> source=<source> [pointTags]
 // Example: "lambda.thumbnail.generate 10 source=thumbnail_service image-format=jpeg"
-sender.SendDeltaCounter("lambda.thumbnail.generate", 10.0, 0, "thumbnail_service", map[string]string{"format" : "jpeg"})
+sender.SendDeltaCounter("lambda.thumbnail.generate", 10.0, "thumbnail_service", map[string]string{"format" : "jpeg"})
 ```
 
 ***Note***: If your `metricName` has a bad character, that character is replaced with a `-`.
--- a/README.md
+++ b/README.md
@@ -224,9 +224,9 @@ Wavefront supports different metric types, such as gauges, counters, delta count
 sender.SendMetric("new-york.power.usage", 42422.0, 0, "go_test", map[string]string{"env" : "test"})
 
 // Wavefront delta counter format
-// <metricName> <metricValue> source=<source> [pointTags]
+// <metricName> <metricValue>  [<timestamp>] source=<source> [pointTags]
 // Example: "lambda.thumbnail.generate 10 source=thumbnail_service image-format=jpeg"
-sender.SendDeltaCounter("lambda.thumbnail.generate", 10.0, "thumbnail_service", map[string]string{"format" : "jpeg"})
+sender.SendDeltaCounter("lambda.thumbnail.generate", 10.0, 0, "thumbnail_service", map[string]string{"format" : "jpeg"})
 ```
 
 ***Note***: If your `metricName` has a bad character, that character is replaced with a `-`.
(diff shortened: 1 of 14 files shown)


joannak-vmware:
Need to fix failed checks.

senders/types.go
```diff
@@ -28,7 +28,7 @@ type MetricSender interface {
 
 	// Sends a delta counter (counter aggregated at the Wavefront service) to Wavefront.
 	// the timestamp for a delta counter is assigned at the server side.
-	SendDeltaCounter(name string, value float64, source string, tags map[string]string) error
+	SendDeltaCounter(name string, value float64, ts int64, source string, tags map[string]string) error
```
vikramraman, line 31: Changing this signature will break backwards compatibility for existing users. Is there a need for a timestamp on delta counters?

README.md
```diff
@@ -224,9 +224,9 @@ Wavefront supports different metric types, such as gauges, counters, delta count
 sender.SendMetric("new-york.power.usage", 42422.0, 0, "go_test", map[string]string{"env" : "test"})
 
 // Wavefront delta counter format
-// <metricName> <metricValue> source=<source> [pointTags]
+// <metricName> <metricValue>  [<timestamp>] source=<source> [pointTags]
 // Example: "lambda.thumbnail.generate 10 source=thumbnail_service image-format=jpeg"
-sender.SendDeltaCounter("lambda.thumbnail.generate", 10.0, "thumbnail_service", map[string]string{"format" : "jpeg"})
+sender.SendDeltaCounter("lambda.thumbnail.generate", 10.0, 0, "thumbnail_service", map[string]string{"format" : "jpeg"})
```
vikramraman, line 229: May need to revert this if we retain the old method signature.

internal/registry.go
```diff
@@ -9,6 +9,10 @@ import (
 type internalSender interface {
 	// Sends a single metric to Wavefront with optional timestamp and tags.
 	SendMetric(name string, value float64, ts int64, source string, tags map[string]string) error
+
+	// Sends a delta counter (counter aggregated at the Wavefront service) to Wavefront.
+	// the timestamp for a delta counter is assigned at the server side.
+	SendDeltaCounter(name string, value float64, ts int64, source string, tags map[string]string) error
```
vikramraman, line 15: Same question here as well. Can omit `ts` since as the comment suggests, the timestamp for a delta counter is assigned at the server side.

senders/formatter.go
```diff
@@ -333,6 +333,11 @@ func sanitizeInternal(str string) string {
 		sb.WriteString(internal.AltDeltaPrefix)
 		skipHead = 2
 	}
+	if (strings.HasPrefix(str, internal.DeltaPrefix) || strings.HasPrefix(str, internal.AltDeltaPrefix)) &&
```
vikramraman, line 336: For clarity, let's add a comment here describing what these lines do

senders/direct.go
```diff
@@ -21,6 +21,22 @@ type directSender struct {
...
+	histogramsInvalid *internal.DeltaCounter
+	histogramsDropped *internal.DeltaCounter
+
+	spansValid   *internal.DeltaCounter
+	spansInvalid *internal.DeltaCounter
+	spansDropped *internal.DeltaCounter
+
+	spanLogsValid   *internal.DeltaCounter
+	spanLogsInvalid *internal.DeltaCounter
+	spanLogsDropped *internal.DeltaCounter
```
vikramraman, line 39: missing internal metrics for events?

vikramraman (changes requested):
Left a few comments. Please take a look. Thanks!

senders/types.go
joannak-vmware: Addd ts for parity with Java and Python SDK (used overloading to give customers the option to add timestamp). Since Go has no overloading, will remove ts parameter.

internal/proxy.go
```diff
@@ -11,23 +11,31 @@ import (
...
+
+	writeSuccesses *DeltaCounter
+	writeErrors    *DeltaCounter
 }
 
-func NewProxyConnectionHandler(address string, flushInterval time.Duration) ConnectionHandler {
-	return &ProxyConnectionHandler{
-		address:     address,
-		flushTicker: time.NewTicker(flushInterval),
+func NewProxyConnectionHandler(address string, flushInterval time.Duration, prefix string, internalRegistry *MetricRegistry) ConnectionHandler {
```
vikramraman, line 30: Couple of points here:
* This is an externally facing constructor. Changing the signature will break backwards compat.
* But also the internal registry should be an internal thing (users should not need to pass it in?). We should default to a prefix and create the internal registry, but allow users to disable this (as described here: https://github.com/wavefrontHQ/wavefront-sdk-go/issues/35)

vikramraman: Nevermind, ignore my previous comment. It's probably okay since this is internal.

vikramraman (approved):

PR #62, opened 2021-02-06
Update version to v0.9.6

joannak-vmware:


open_source_licenses.txt
```diff
@@ -1,6 +1,6 @@
 open_source_licenses.txt
 
-Wavefront by VMware SDK for Go 0.9.5 GA
+Wavefront by VMware SDK for Go 0.9.6 GA
```
vikramraman, line 3: Should this be 0.9.7 since that is the currently released version?

joannak-vmware: I see, will update correctly. Thanks for catching!

vikramraman (approved):

PR #65, opened 2021-06-21
Add a no-op WavefrontSender implementation

mssrinivas:


oppegard (changes requested):
Thanks for the submission @mssrinivas! Could you give an example of your use-case that requires a no-op Sender, rather than just not instantiating a Sender at all?

mssrinivas:
@oppegard  We have a service A currently using DirectConfig to send metrics to WF. We also have another service B which implements interfaces of service A but do not want to enable metrics capturing. Since things are tightly coupled we wanted to have a way to initialize Noop sender.

oppegard (approved):

hanwavefront (approved):
LGTM, thanks!

oppegard (approved):

PR #67, opened 2021-07-02
Optimize formatter performance with StringBuilder mod

fredwangwang:
The Go standard StringBuilder has some limitations, this pr included a modified version of StringBuilder. The changes include: 

- do not free the underlying buffer when Reset(); so it can be reused in pool
- add AppendFloat and AppendInt methods; to avoid additional alloc caused by strconv.Format*
- add Bytes() method

Besides the mod above also include:

- cleanup unnecessary conditions in sanitize
- remove the additional allocations used by sanitize

## Benchmark Result

cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz

Original:

```
BenchmarkMetricLine-12 875883 1229 ns/op 250 B/op 17 allocs/op
BenchmarkHistoLine-12 635227 1729 ns/op 614 B/op 22 allocs/op
BenchmarkSpanLine-12 1623225 747.5 ns/op 340 B/op 11 allocs/op
```

Optimized:

```
BenchmarkMetricLine-12 3926373 309.9 ns/op 64 B/op 1 allocs/op
BenchmarkHistoLine-12 1764009 674.7 ns/op 416 B/op 6 allocs/op
BenchmarkSpanLine-12 4476735 266.7 ns/op 192 B/op 1 allocs/op
```

Summary:

| Method     | CPU Speedup | Alloc Reduction |
|------------|-------------|-----------------|
| MetricLine | 4.0x        | -94%            |
| HistoLine  | 2.5x        | -72%            |
| SpanLine   | 2.8x        | -91%            |


idrennanvmware:
@fredwangwang ty for taking this on!!

fredwangwang:
failure on testing against go 1.11 for missing `strings.ReplaceAll`. Is that version (almost 3 years ago) still need to be supported?

keep94:
Closing this PR.  Since this is an API that makes network calls, most of the time will be spent doing HTTP requests. It seems like tweaking StringBuilder would be premature optimization.  However, if you are able to justify the need for this optimization as well as provide test coverage for the modified StringBuilder, we would be happy to review this once again. Thank you for contributing.

PR #69, opened 2021-07-21
Support sending spans and metrics to different ports using the WavefrontSender

thepeterstone:
This PR adds options for the constructor for `WavefrontSender` to specify tracing port and metrics port separately from the endpoint hostname so that metrics can be sent to the correct port even when using a wavefront-proxy's `customTracingListenerPorts`.


oppegard (approved):
lgtm

PR #75, opened 2022-01-21
Update go version, which forced me to run go mod tidy..

johncornish:
I'm not sure why upgrading the go version required these changes as forced by `go mod tidy`, but happy to get your feedback


thepeterstone (approved):
I think the go.mod rearrangement is expected when updating go version, looks good to me

PR #76, opened 2022-02-04
Allow for tagging internal SDK metrics

deosu:
Update newProxySender to take an additional setter with Otel tag set by
the Otel-Collector.

### DIFF
--- a/internal/registry.go
+++ b/internal/registry.go
@@ -38,7 +38,7 @@ func SetSource(source string) RegistryOption {
 
 func SetInterval(interval int) RegistryOption {
 	return func(registry *MetricRegistry) {
-		registry.reportTicker = time.NewTicker(time.Second * time.Duration(interval))
+		registry.reportTicker = time.NewTicker(time.Millisecond * time.Duration(interval))
 	}
 }
 
--- a/internal/registry.go
+++ b/internal/registry.go
@@ -19,7 +19,7 @@ type internalSender interface {
 type MetricRegistry struct {
 	source       string
 	prefix       string
-	Tags         map[string]string
+	tags         map[string]string
 	reportTicker *time.Ticker
 	sender       internalSender
 	done         chan struct{}
@@ -44,16 +44,16 @@ func SetInterval(interval int) RegistryOption {
 
 func SetTags(tags map[string]string) RegistryOption {
 	return func(registry *MetricRegistry) {
-		registry.Tags = tags
+		registry.tags = tags
 	}
 }
 
 func SetTag(key, value string) RegistryOption {
 	return func(registry *MetricRegistry) {
-		if registry.Tags == nil {
-			registry.Tags = make(map[string]string)
+		if registry.tags == nil {
+			registry.tags = make(map[string]string)
 		}
-		registry.Tags[key] = value
+		registry.tags[key] = value
 	}
 }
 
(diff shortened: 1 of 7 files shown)


senders/proxy.go
```diff
@@ -133,6 +135,15 @@ func makeConnHandler(host string, port, flushIntervalSeconds int, prefix string,
 	return internal.NewProxyConnectionHandler(addr, flushInterval, prefix, internalRegistry)
 }
 
+func SetTag(key, value string) internal.RegistryOption {
```
oppegard, line 138: We need to communicate to SDK users that these tags only apply to internal SDK metrics. I suggest we try just adding a field to ProxyConfiguration of `SDKMetricsTags map[string]string`. I think that would also allow rolling back the change of making `MetricRegistry.Tags` public.

oppegard: Also, can you look at adding the same feature to the sender in `client.go`? That's our non-deprecated sender, and I'd like to keep feature parity.

deosu: Sure, thanks @oppegard

senders/client_factory.go
```diff
@@ -174,3 +183,10 @@ func TracesPort(port int) Option {
 		cfg.TracesPort = port
 	}
 }
+
+// SDKMetricsTags sets internal SDK metrics.
+func SDKMetricsTags(tags map[string]string) Option {
+	return func(cfg *configuration) {
+		cfg.SDKMetricsTags = tags
```
keep94, line 190: Just to be safe, consider making a defensive copy of tags here.  That way the Option SDKMetricsTags() creates won't change even if the caller changes the map they passed in later on.

senders/client_test.go
```diff
@@ -58,6 +58,13 @@ func TestSendDirect(t *testing.T) {
 	doTest(t, wf)
 }
 
+func TestSendDirectWithTags(t *testing.T) {
+	tags := map[string]string{"foo": "bar"}
+	wf, err := senders.NewSender("http://"+token+"@localhost:"+wfPort, senders.SDKMetricsTags(tags))
```
keep94, line 63: Just curious what this is testing.  This line uses "foo": "bar" as the SDK metric tags, but I don't see where this is verified.  I mean you could leave `senders.SDKMetricsTags(tags)` out and the test would still pass.

senders/proxy_test.go
```diff
@@ -22,26 +21,67 @@ func netcat(addr string, keepopen bool) {
 	lis.Close()
 }
 
-func init() {
+func TestProxySends(t *testing.T) {
 	go netcat("localhost:30000", false)
 	go netcat("localhost:40000", false)
 	go netcat("localhost:50000", false)
+
+	time.Sleep(time.Second)
```
keep94, line 29: Consider adding a TODO here to see if there is a better way to avoid the race condition. I asked on the golang group, but waiting for a response.

keep94 (changes requested):
Thanks for doing this work Sumit.  I went ahead and approved the changes since I will be out next week. That way I don't block you. However, I did have a few suggested changes for you, so please have a look at my comments. Have a great week.

keep94 (approved):

senders/client_test.go
deosu: Travis, I wanted to make sure that the WFSender is still able to send spans & metrics after we add the SDKMetricsTag, But you are correct, the unit test is missing an assertion to check if these SDKMetricsTags are set. I am looking into options to assert it.

senders/proxy_test.go
deosu: Add TODO

oppegard (changes requested):

internal/registry.go
```diff
@@ -38,7 +38,7 @@ func SetSource(source string) RegistryOption {
 
 func SetInterval(interval int) RegistryOption {
```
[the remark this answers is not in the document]
deosu: changed

internal/registry.go
```diff
@@ -38,7 +38,7 @@ func SetSource(source string) RegistryOption {
...
 	return func(registry *MetricRegistry) {
-		registry.reportTicker = time.NewTicker(time.Second * time.Duration(interval))
+		registry.reportTicker = time.NewTicker(time.Millisecond * time.Duration(interval))
```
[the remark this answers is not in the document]
deosu: changed

oppegard (approved):
Looks good!

PR #79, opened 2022-02-10
feat: Bump version to 0.9.10

deosu:
Bump version to 0.9.10


oppegard (approved):
LGTM. In the future, I'd like to see `feat: ` removed from commit messages. I don't feel it adds much value when scanning the git log.

PR #81, opened 2022-04-05
Improve error messages when formatting telemetry

oppegard:


senders/formatter.go
```diff
@@ -125,18 +125,18 @@ func HistoLine(name string, centroids histogram.Centroids, hgs map[histogram.Gra
...
+		return "", errors.New("span name cannot be empty")
 	}
 
 	if source == "" {
 		source = defaultSource
 	}
 
 	if !isUUIDFormat(traceId) {
-		return "", errors.New("traceId is not in UUID format")
+		return "", fmt.Errorf("traceId is not in UUID format: span=%s", name)
```
keep94, line 136: Would it be worth showing the trace ID in the error?

senders/formatter.go
```diff
@@ -125,18 +125,18 @@ func HistoLine(name string, centroids histogram.Centroids, hgs map[histogram.Gra
...
 	}
 	if !isUUIDFormat(spanId) {
-		return "", errors.New("spanId is not in UUID format")
+		return "", fmt.Errorf("spanId is not in UUID format: span=%s", name)
```
keep94, line 139: Same here, should we show span ID?

senders/formatter_test.go
```diff
@@ -192,6 +194,32 @@ func TestSpanLine(t *testing.T) {
...
+	_, err = SpanLine("a_name", 0, 0, "00-00", "", uuid, nil, nil, nil, nil, "")
+	require.Error(t, err)
+	assert.Equal(t, "traceId is not in UUID format: span=a_name", err.Error())
+
+	_, err = SpanLine("a_name", 0, 0, "00-00", uuid, "", nil, nil, nil, nil, "")
+	require.Error(t, err)
+	assert.Equal(t, "spanId is not in UUID format: span=a_name", err.Error())
+
+	_, err = SpanLine("a_name", 0, 0, "a_source", uuid, uuid, nil, nil,
+		[]SpanTag{{Key: "", Value: ""}}, nil, "")
```
keep94, line 213: Nit: Struct fields in GO get their zero values if left uninitialized. So you don't have to explicitly set Value to the empty string here.  ```[]SpanTag{{Key: ""}}``` is enough since you are testing what happens if Key is empty.

keep94 (approved):
Thanks for doing this work Glenn. I am approving as is, but I did have some questions / comments so read on.

senders/formatter_test.go
oppegard: Hm, I'm not sure how I feel about that. I think that "future-me" will better understand the intent of what's being tested by explicitly setting `Value` to empty string.

senders/formatter.go
oppegard: Good idea – I added it.

senders/formatter.go
oppegard: Good idea – I added it.

Issue #84, opened 2022-06-15
senders.NewSender doesn't work for direct ingestion


keep94:
If you call 
```
sender, err := senders.NewSender("https://11111111-2222-3333-4444-555555555555@surf.wavefront.com")
```
To get a sender that does direct ingestion to https://surf.wavefront.com, it won't work because it tries to send to port 2878 instead of to port 443.

If you call
```
sender, err := senders.NewSender("https://11111111-2222-3333-4444-555555555555@surf.wavefront.com:443")
```

This does work, but a user of the API will not think to do this.

This API should be intuitive for users.  Both chunks of code ought to work.

LukeWinikates:
@keep94 does it make sense to close this issue since it has a merged PR tied to it?

oppegard:
Pinging @keep94: was this fixed in #85, so this issue should be closed?

PR #85, opened 2022-06-15
Fix senders.NewSender() for direct ingestion.

keep94:
This commit fixes Issue #84.

### DIFF
--- a/senders/client_factory.go
+++ b/senders/client_factory.go
@@ -44,6 +44,11 @@ type configuration struct {
 	SDKMetricsTags       map[string]string
 }
 
+func (c *configuration) setDefaultPort(port int) {
+	c.MetricsPort = port
+	c.TracesPort = port
+}
+
 // NewSender creates Wavefront client
 func NewSender(wfURL string, setters ...Option) (Sender, error) {
 	cfg, err := CreateConfig(wfURL, setters...)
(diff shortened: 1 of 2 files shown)


senders/client_factory.go
```diff
@@ -67,15 +67,26 @@ func CreateConfig(wfURL string, setters ...Option) (*configuration, error) {
...
 	if len(u.User.String()) > 0 {
 		cfg.Token = u.User.String()
 		u.User = nil
 	}
 
+	switch strings.ToLower(u.Scheme) {
+	case "http":
+		if cfg.Token != "" {
+			cfg.MetricsPort = 80
+			cfg.TracesPort = 80
```
oppegard, line 79: There are four places in this method we set `MetricsPort` and `TracesPort` (some of the conditionally). Can we DRY up setting the defaults via helper function, where it determines if it's DI or Proxy and HTTP or HTTPS?

`func defaultTelemetryPorts(...) (metricsPort, tracesPort int)`

oppegard (changes requested):

senders/client_factory.go
keep94: I added a method to configuration to set the default port.

oppegard (approved):

Issue #86, opened 2022-06-15
senders.SDKMetricsTags() creates an Option that can change in unexpected ways.


keep94:
Consider these code snippets

```
map1 := map[string]string{"foo": "bar"}
option1 := senders.SDKMetricsTags(map1)
map1["foo"] = "wrong"

// option1 now represents adding "foo=wrong" tag, not "foo=bar"
sender, err := senders.NewSender("http://localhost", option1)
```

```
option1 := senders.SDKMetricsTags(map[string]string{"foo": "bar"})
option2 := senders.SDKMetricsTags(map[string]string{"baz": "none"})
sender, err := senders.NewSender("http://localhost", option1, option2)

// Now option1 represents adding both foo=bar and baz=none. badSender won't behave
// as user expects.
badSender, err := senders.NewSender("http://10.1.10.1", option1)
```

options in the API shouldn't change once created. User of API should be able to reuse a created option at any time knowing that it represents the same thing.

PR #87, opened 2022-06-15
Ensure Option instances are immutable.

keep94:
This fixes issue #86.


senders/client_factory.go
```diff
@@ -185,14 +185,25 @@ func TracesPort(port int) Option {
 
 // SDKMetricsTags sets internal SDK metrics.
 func SDKMetricsTags(tags map[string]string) Option {
+	// prevent caller from accidentally mutating this option.
+	tags = copyTags(tags)
```
deosu, line 189: Can we rename this variable if the same name `tags` is confusing? May be `copiedTags` or something else?

deosu, line 189: Can we rename this variable if the same name `tags` is confusing? May be `copiedTags` or something else?

deosu (approved):
LGTM

senders/client_factory.go
keep94: Done. tags -> copiedTags.

PR #96, opened 2022-08-17
In new Go API, name internal metrics based on sender mode.

keep94:


senders/client_factory.go
```diff
@@ -2,11 +2,12 @@ package senders
 
 import (
 	"fmt"
-	"github.com/wavefronthq/wavefront-sdk-go/internal"
 	"net/url"
 	"os"
 	"strconv"
 	"strings"
+
+	"github.com/wavefronthq/wavefront-sdk-go/internal"
```
deosu, line 10: Can we revert this change?

senders/client_factory.go
```diff
@@ -44,6 +45,18 @@ type configuration struct {
 	SDKMetricsTags       map[string]string
 }
 
+func (c *configuration) Direct() bool {
```
deosu, line 48: I am not sure about the naming convention that we follow, but usually, in java, we use something like `isDirectIngested`.

deosu (approved):
Just a few small comments, but otherwise LGTM.

senders/client_factory.go
keep94: Generally with go, standard library imports are listed first followed by a space and then 3rd party imports.

senders/client_factory.go
keep94: Thanks for catching this. Although it is allowed to write methods like GetXXX, it usually isn't done that way in GO.  See https://go.dev/doc/effective_go#Getters.  I am assuming that not having GetXXX methods in Go carries over to not having IsXXX methods in Go as well.

Issue #98, opened 2022-08-19
Is it time to declare `v1` and remove the old factory methods?


LukeWinikates:
The SDK version is currently `0.10.2`, but the SDK is in active use in production environments. Is now the right time to declare a v1?

Also, since the SDK is not yet v1, how will we decide when it is ok to remove the code for the old factory methods and the TCP connection? If we are not actively using these ourselves anymore, maybe it's time to take them out and not carry the burden of the keeping the deprecated functions working.

keep94:
Luke, your idea of parallel http requests sounds interesting. I just wonder how that will work with the Flush() function of a sender object.  When you create a sender, you can effectively turn off autoflushing, send a bunch of metrics and spans which get buffered and flush manually.  Flush() returns an error indicating whether or not everything buffered up was sent successfully.  This sort of use of the API implies just one goroutine using the sender, so I am wondering how a pool of http connections would work.  Are you possibly thinking of rewriting flush to use multiple http connections to do the flushing?

Regarding releasing a v1, I feel that there are several artifacts in the public API that shouldn't be exposed. For instance, I think the functions that convert metrics to strings to be sent over the wire could be made private.


Back to the parallel http requests, if Flush() were rewritten to use multiple http connections, it would still have to block until that last http connection finished sending its portion of metrics/traces because by contract Flush() has to return an error indicating whether or not everything was successfully sent.

Issue #89, opened 2022-06-16
Audit public SDK functions to see if they need to be public for next major release


keep94:
There are several methods in the senders package that show up on pkg.go.dev but were probably not intended to be used by clients of the API.  We should identify such methods so that they can be made private in the next major release.

PR #100, opened 2022-08-20
hide a few functions that do not need to be part of the public API

LukeWinikates:
related to issues #89 and #98

### DIFF
--- a/senders/client.go
+++ b/senders/client.go
@@ -75,7 +75,7 @@ func (sender *wavefrontSender) Start() {
 }
 
 func (sender *wavefrontSender) SendMetric(name string, value float64, ts int64, source string, tags map[string]string) error {
-	line, err := MetricLine(name, value, ts, source, tags, sender.defaultSource)
+	line, err := metricLine(name, value, ts, source, tags, sender.defaultSource)
 	if err != nil {
 		sender.pointsInvalid.Inc()
 		return err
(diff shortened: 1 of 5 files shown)


LukeWinikates:
@keep94 - not sure if this is exactly what you had in mind, and maybe you had your own reasons for not wanting to hide these methods right away. I figured a PR would give us some code to talk about.

keep94:
Luke. So sorry, I am out on vacation this week. I'll be back Friday
morning. I will have a look at your PR then. Thank you for doing this work.

keep94 (commented):
Luke. Thanks for the work. This is a good start.  One question, why didn't you change the HistoLine() function to histoLine()?  

Before committing these changes and going to v1.0.0, we should loop in @oppegard my tech lead to see if he wants anything in the API changed before moving to v1.

LukeWinikates:
> Luke. Thanks for the work. This is a good start. One question, why didn't you change the HistoLine() function to histoLine()?

That's just an oversight :) I'll fix it now

> Before committing these changes and going to v1.0.0, we should loop in @oppegard my tech lead to see if he wants anything in the API changed before moving to v1.

Sounds good to me. I would also like to go ahead and delete the old factory methods prior to v1.

LukeWinikates:
@keep94 - I hid those methods and also changed `histo` to `histogram` in several places. Thanks for taking a look at this!

oppegard (approved):
Thanks!

PR #104, opened 2022-09-02
Automatically pull version ID from build.

keep94:
Add version tag to internal metrics.

### DIFF
--- a/senders/client_factory.go
+++ b/senders/client_factory.go
@@ -213,11 +213,12 @@ func TracesPort(port int) Option {
 	}
 }
 
-// SDKMetricsTags adds the tags provided in tags to all internal metrics
-// this library reports. Clients can use multiple SDKMetricsTags calls when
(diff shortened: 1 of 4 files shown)


senders/client_factory.go
```diff
@@ -219,14 +222,8 @@ func SDKMetricsTags(tags map[string]string) Option {
 	// prevent caller from accidentally mutating this option.
```
oppegard, line 222: Can you update the documentation for this function? It says:

> By default, the sender does not attach any tags to internal metrics.

But it attaches a `pid` and `version` tag by default now.

oppegard (approved):
I requested an update to a function comment. Otherwise, lgtm.

senders/client_factory.go
keep94: Nice catch.  Done.

oppegard (approved):

Issue #108, opened 2022-10-07
bug: senders.NewSender does not support urls with both port and path.


LukeWinikates:
This was originally reported as a bug on the wavefront telegraf plugin: https://github.com/influxdata/telegraf/issues/11925

`senders.CreateConfig("http://localhost:8071/wavefront")` currently removes the `port` portion of the url during the parsing process. This doesn't create any problems for urls without a path component, but when port and path are both present, the resulting url is mangled: `"http://localhost/wavefront"`, and the port is then appended by the reporter later, resulting in something like `"http://localhost/wavefront:8071/report?f=wavefront"`.

I'm hoping to submit a PR with test cases in the next couple of hours.

PR #109, opened 2022-10-07
bug: address parsing wavefront urls with a path component

LukeWinikates:
resolves: #108

This PR adds a "Path" field to the `configuration` struct so that we can stash the path component of a URL there and use it to correctly reconstruct a URL later.

I would be interested in suggestions about the `MetricsURL` and `TracesURL` methods; I think exporting these is a little bit unfortunate, but the url construction is important to test and is currently a little hard to exercise in a test because it's happening inside unexported functions. We could refactor to make this easier to test, but I didn't want to increase the scope of the PR for this fix.

This supports usecases where rather than talking to a WF instances or proxy directly, we are communicating through some sort of proxy that does path-based routing of our requests.

@keep94 @oppegard thanks in advance for your eyes on this PR 🙇‍♂️

cc: @rushikeshbgithub

### DIFF
--- a/internal/reporter.go
+++ b/internal/reporter.go
@@ -32,12 +32,12 @@ func (reporter reporter) Report(format string, pointLines string) (*http.Respons
 		return nil, formatError
 	}
 
-	buf, err := linesToBuffer(pointLines)
+	requestBody, err := linesToGzippedBytes(pointLines)
 	if err != nil {
 		return &http.Response{}, err
 	}
 
-	req, err := reporter.buildRequest(format, buf)
+	req, err := reporter.buildRequest(format, requestBody)
 	if err != nil {
 		return &http.Response{}, err
 	}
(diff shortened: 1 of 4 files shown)


internal/reporter.go
```diff
@@ -32,22 +32,38 @@ func (reporter reporter) Report(format string, pointLines string) (*http.Respons
...
+
+	req, err := reporter.buildRequest(format, buf)
+	if err != nil {
+		return &http.Response{}, err
+	}
+
+	return reporter.execute(req)
+}
+
+func linesToBuffer(pointLines string) (bytes.Buffer, error) {
```
keep94, line 48: I don't think bytes.Buffer supports copying by simple assignment. Instead of returning a `bytes.Buffer` or even a `*bytes.Buffer` consider instead returning a `[]byte`.  Returning a `[]byte` is more concise and less confusing because you are eagerly constructing the bytes in this function anyway.  Also consider renaming `linesToBuffer` to `linesToBytes` or something like that.

internal/reporter.go
```diff
@@ -32,22 +32,38 @@ func (reporter reporter) Report(format string, pointLines string) (*http.Respons
...
+		return bytes.Buffer{}, err
 	}
 	if err = zw.Close(); err != nil {
-		return nil, err
+		return bytes.Buffer{}, err
 	}
+	return buf, err
+}
 
+func (reporter reporter) buildRequest(format string, buf bytes.Buffer) (*http.Request, error) {
```
keep94, line 62: Instead of passing in a bytes.Buffer, pass in a simple `[]byte` instead. This makes buildRequest() easier to use.

internal/reporter.go
```diff
@@ -32,22 +32,38 @@ func (reporter reporter) Report(format string, pointLines string) (*http.Respons
...
 	apiURL := reporter.serverURL + reportEndpoint
 	req, err := http.NewRequest("POST", apiURL, &buf)
```
keep94, line 64: Here you can simply create a `bytes.Buffer` from you `[]byte` slice on the fly using `NewBuffer` and the right things will happen.

senders/client_factory.go
```diff
@@ -142,6 +148,14 @@ func newWavefrontClient(cfg *configuration) (Sender, error) {
 	return sender, nil
 }
 
+func TracesURL(cfg *configuration) string {
```
keep94, line 151: Could we instead make TracesURL() and MetricsURL() methods of `configuration`?  I think you can still test after making this change.

keep94 (changes requested):
Thanks for doing this work LukeWinikates.  I just have a few suggestions.

internal/reporter.go
LukeWinikates: I think what you're saying is that returning a `bytes.Buffer` is a form of copying by value and that the buffer might not end up containing the intended gzipped content. I haven't dug into that but I like the idea of returning the simpler `[]byte` type - pushing that change now.

LukeWinikates: Slightly concerned about whether the refactor, although more readable, might cause any noticeable performance changes given we're creating a new buffer twice now (I don't know if we think of this code as performance sensitive or have any performance tests around it)

keep94 (approved):
Looks Good To Me

oppegard (approved):

PR #110, opened 2022-10-11
testing: add a basic integration test

LukeWinikates:
In the process of working on #109 , I was worried about not having automated end-to-end test coverage, and I thought that even a simple test against a local http server might make it easier to refactor the sdk internals with greater confidence.

@keep94 what do you think?

cc: @suprajanarasimhan @ktollivervmw

### DIFF
--- a/senders/test_server_test.go
+++ b/senders/test_server_test.go
@@ -1,0 +1,54 @@
+package senders
+
+import (
+	"bufio"
+	"compress/gzip"
+	"net/http"
+	"net/http/httptest"
+)
+
+func startTestServer() *testServer {
+	handler := &testServer{}
+	server := httptest.NewServer(handler)
+	handler.httpServer = server
+	handler.URL = server.URL
+	return handler
+
+}
+
+type testServer struct {
+	MetricLines []string
+	httpServer  *httptest.Server
+	URL         string
+}
+
+func (s *testServer) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
+	newLines, err := decodeMetricLines(request)
+	if err != nil {
+		writer.WriteHeader(500)
+	}
+	s.MetricLines = append(s.MetricLines, newLines...)
+	writer.WriteHeader(200)
+}
+
+func decodeMetricLines(request *http.Request) ([]string, error) {
+	var metricLines []string
+	reader, err := gzip.NewReader(request.Body)
+	if err != nil {
+		return metricLines, err
+	}
+	scanner := bufio.NewScanner(reader)
+	for scanner.Scan() {
+		line := scanner.Text()
+		metricLines = append(metricLines, line)
+	}
+	if scanner.Err() != nil {
+		reader.Close()
+		return metricLines, scanner.Err()
+	}
+	return metricLines, reader.Close()
+}
+
+func (s *testServer) Close() {
+	s.httpServer.Close()
+}
--- a/senders/integration_test.go
+++ b/senders/integration_test.go
@@ -1,0 +1,19 @@
+package senders
+
+import (
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestEndToEnd(t *testing.T) {
+	testServer := startTestServer()
+	defer testServer.Close()
+	sender, err := NewSender(testServer.URL)
+	require.NoError(t, err)
+	require.NoError(t, sender.SendMetric("my metric", 20, 0, "localhost", nil))
+	require.NoError(t, sender.Flush())
+
+	assert.Equal(t, 1, len(testServer.MetricLines))
+	assert.Equal(t, "\"my-metric\" 20 source=\"localhost\"", testServer.MetricLines[0])
+}


senders/test_server_test.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"bufio"
+	"compress/gzip"
+	"context"
+	"fmt"
+	"net/http"
+)
+
+func newTestServer(port int) *testServer {
+	metricLines := make([]string, 0)
```
keep94, line 12: You can shorten this function to just be
```
return &testServer{port: port}
```
You can append to a nil slice and the right things will happen.

senders/test_server_test.go
```diff
@@ -0,0 +1,73 @@
...
+		reader.Close()
+	}
+	scanner := bufio.NewScanner(reader)
+
+	for scanner.Scan() {
+		line := scanner.Text()
+		metricLines = append(metricLines, line)
+	}
+
+	if err = reader.Close(); err != nil {
```
keep94, line 65: Consider calling reader.Close() in a defer statement.  Once reader is successfully created, you have to close it when done anyway.

keep94 (changes requested):
Thanks for adding this integration test, LukeWinikates.  I'd like to see this test use the httptest.Server type which binds to a random opened port.  Recently I had trouble running a test suite of a library because I was already running a service on the same port it happened to be trying to use.

senders/test_server_test.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+		writer.WriteHeader(200)
+	}
+}
+
+func decodeMetricLines(request *http.Request) ([]string, error) {
+	metricLines := make([]string, 0)
+	reader, err := gzip.NewReader(request.Body)
+	if err != nil {
+		reader.Close()
```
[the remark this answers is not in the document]
LukeWinikates: Yeah, certainly looking at the way `gzip.NewReader` is implemented now, this would be nil. I also realized that this should return early with the error, so I'll fix that.

senders/test_server_test.go
LukeWinikates: This looked a little funny to me as well - it's modified from the gzip logic that is used to compress the lines in the reporter, so this pattern shows up there as well.

I think that `defer` would do the right thing in terms of making sure that the reader is always closed without having to call it in two different spots, but it would also mean that we couldn't return the `error` from `reader.Close`, which I think could cause us to drop some useful error messages, at least in theory. I've been inclined to keep using this pattern so that we're consistently returning all the error values we get.

I changed this up a little bit, but if you think there's a better way to use defer and still return the error, I'd definitely be up for trying it out.

senders/test_server_test.go
```diff
@@ -0,0 +1,52 @@
...
+	httpServer  *httptest.Server
+	URL         string
+}
+
+func (s *testServer) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
+	newLines, err := decodeMetricLines(request)
+	if err != nil {
+		writer.WriteHeader(500)
+	}
+	for _, metricLine := range newLines {
```
keep94, line 30: You can just say ```s.MetricLines = append(s.MetricLines, newLines...)```

senders/test_server_test.go
```diff
@@ -0,0 +1,52 @@
...
+		s.MetricLines = append(s.MetricLines, metricLine)
+	}
+	writer.WriteHeader(200)
+}
+
+func decodeMetricLines(request *http.Request) ([]string, error) {
+	metricLines := make([]string, 0)
```
keep94, line 37: You can just say ```var metricLines []string```

senders/test_server_test.go
```diff
@@ -0,0 +1,52 @@
...
+	reader, err := gzip.NewReader(request.Body)
+	if err != nil {
+		return metricLines, err
+	}
+	scanner := bufio.NewScanner(reader)
+	for scanner.Scan() {
```
keep94, line 43: scanner can return errors through scanner.Err().  Have a look at https://pkg.go.dev/bufio#example-Scanner-Custom to see how.

keep94 (commented):
Thanks LukeWinikates for this work.  Just a few minor comments.

senders/test_server_test.go
LukeWinikates: ah! I thought there was a splat syntax but tried `...newLines` and when that didn't work I just rolled with the for loop 🤣

senders/test_server_test.go
```diff
@@ -0,0 +1,53 @@
...
+		line := scanner.Text()
+		metricLines = append(metricLines, line)
+	}
+	if scanner.Err() != nil {
+		return metricLines, scanner.Err()
```
keep94, line 46: If you get here, you will leak the gzip.Reader

keep94 (commented):
Just one comment.

senders/test_server_test.go
LukeWinikates: Thanks for the catch! I just pushed a change adding `reader.Close()` right before this line. That seems like a reasonable choice but it did have me wishing I could just use `defer`.

keep94 (approved):
LTGM

PR #111, opened 2022-10-12
[#183035188] Make request timeout and TLS properties configurable

suprajanarasimhan:
Related to #6. 
Creating this draft in case there is any early feedback to act on. 

Before this DRAFT can be finalized on our team's end, the following things are needed: 
- [x] Consider refactoring NewReporter() to take in the entire struct of type `configuration` to avoid passing multiple parameters to it. 
- [x] Invoke NewSender() and exercise TLS manually (i.e. outside of the test suite) to validate.
- [x] Updating the README.md to reference examples of invoking newly introduced timeout and TLS-related setters. 
- [x] Update commit message to remove our team's issue tracking ID. 
- [Won't fix] Fix git history (1) To eliminate merge commits (2) To squash all changes into a single commit. -- See how [here](https://medium.com/@topspinj/how-to-git-rebase-into-a-forked-repo-c9f05e821c8a) and do it on branch `expose-tls-properties`.  -- Created a fresh branch instead.


keep94:
Please resolve merge conflicts.

senders/client_factory.go
```diff
@@ -44,6 +45,10 @@ type configuration struct {
 	// together with batch size controls the max theoretical throughput of the sender.
 	FlushIntervalSeconds int
 	SDKMetricsTags       map[string]string
+
+	Timeout int
```
oppegard, line 49: Is there any reason not to make this a `time.Duration`, matching the type that's used in `http.Client{}`? It certainly seems more expressive. Alternatively, name the field something like `TimeoutInSeconds`.

LukeWinikates:
@suprajanarasimhan there's now an example of an end-to-end integration test in `senders/integration_test.go` - you should be able to extend that pattern to introduce an end-to-end tls configuration test. The current test is just http, but the same httptest package includes a method for starting a TLS server: https://pkg.go.dev/net/http/httptest#NewTLSServer.

senders/client_factory.go
suprajanarasimhan: Hi @oppegard, thanks for the feedback! Of the 2 options, I am leaning toward renaming the field to `TimeoutInSeconds`. 

Why: The caller of NewSender() can invoke a `TimeoutInSeconds()` setter with a `int` (rather than having to think about constructing `time.Duration`).  Then all they have to think about will be how long they want the timeout to be. 

However I do understand that this choice is inconsistent with the following theme: Reuse types used in `http.Client`. Especially because the TLSConfigOptions() exemplifies this theme. 

I'll check with `@LukeWinikates` before finalizing this change.

LukeWinikates: @suprajanarasimhan I would say go with `time.Duration`. I agree it's slightly less convenient to create than an integer literal, but typically you would do something like `15 * time.Seconds` to create a `time.Duration`, so it ends up being very readable once you know about the pattern.

suprajanarasimhan: Thank you!

suprajanarasimhan:
Hi @oppegard, thank you for your advice on this PR!

Since I am new to rebasing changes from a main repository onto a fork, I ended up making some undesired merge commits in git history. 

Rather than handle the complexity of fixing this in-place on the current branch, I decided to create a fresh branch and https://github.com/wavefrontHQ/wavefront-sdk-go/pull/122. The downside is that that PR conversation won't be preserved there. The upside is that I have learned a better workflow next time. 

Cc'ing @LukeWinikates in case either of you have feedback for me learn for next time.

PR #107, opened 2022-09-22
docs: simplify README.md

LukeWinikates:
* move releasing instructions into their own file
* remove table discussing SDKs not in this repo
* remove examples in README in favor of code examples
* add internal SDK metrics to this README

PR #113, opened 2022-10-21
This PR is like #107, but with the examples fixed

keep94:


keep94:
This PR was for illustration purposes only. Closing.

PR #115, opened 2022-10-24
Make senders.Sender be a concrete type.

keep94:
Here is what I am thinking about making senders.Sender be a concrete type. This is a breaking change that will cause most clients to have to change their code.

After doing this exercise, I am thinking we probably don't need the DistributionSender, MetricSender, or EventSender interfaces. They are declared in the sender package but they don't seem to be used anywhere. We can get rid of these extra interfaces in another PR.


senders/client.go
```diff
@@ -9,8 +9,86 @@ import (
...
+func (s *Sender) GetFailureCount() int64 {
+	if s.spec == nil {
+		return 0
+	}
+	return s.spec.GetFailureCount()
+}
+
 // Sender Interface for sending metrics, distributions and spans to Wavefront
-type Sender interface {
+type senderSpec interface {
```
LukeWinikates, line 91: Could we give this a different name that describes its function more? I'm not sure what it would be, since its responsibility seems to be "sending". I was thinking about about the `http.Client` object and its internal `Transport`.

LukeWinikates:
@keep94 I just thought of something - I think that eliminating the interface means that `MultiSender` will no longer work, at least not as cleanly as it previously did.

keep94:
We decided to leave Sender as an interface so we merged #116 instead.

PR #116, opened 2022-10-24
Make Sender so that it can't be implemented externally.

keep94:
This is another way to ensure that we can add methods to Sender as non-breaking changes.  This works but it is weird because we declare a 9 method interface in GO, a language where interfaces with just 1 or 2 methods are common. Also, note that there is not a way to write examples for interface methods in GO. I think this is by design.

The up side to this solution is that it won't break most existing clients.

### DIFF
--- a/senders/client.go
+++ b/senders/client.go
@@ -17,6 +17,7 @@ type Sender interface {
 	EventSender
 	internal.Flusher
 	Close()
+	private()
 }
 
 type wavefrontSender struct {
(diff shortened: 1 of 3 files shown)


keep94:
@oppegard I may have been unclear in my description of this PR.  It is fine if we or someone else adds methods to the Sender interface in the future. I anticipate we will need to do that sometime in the future. This PR will allow us to safely add methods to the Sender interface as non breaking changes going forward.  What we don't want is for someone outside of our organization to write their own concrete type that implements Sender and then pass their custom concrete type to a function that accepts a Sender interface.  If they are successful at doing that then as soon as we add any new methods to the Sender interface we would break that person's code so that it doesn't compile. That person would have to fix their code by adding implementations of any new method(s) we added to the Sender interface in their concrete type. This would be a breaking change on our end.  To avoid this type of scenario in GO, it is idiomatic to add private methods to the interface to keep external parties from ever writing a successful implementation of it.  A good example of this is the Type interface in the reflect package of Go.

As far as whether this PR or PR #115 is better I am indifferent.  #115 would allow us to add examples for individual methods of sender, but #116 is a more friendly breaking change.

LukeWinikates (approved):

oppegard (approved):
LGTM

PR #117, opened 2022-10-25
Make senders.Sender be a concrete value type.

keep94:
This PR makes Sender a concrete type but uses value semantics instead of pointer semantics. This too is weird in that it doesn't make sense to assign one sender to another as you would assign a time variable to another. But the advantage of this PR is that it would cause less breakage than doing it the right way (using pointer semantics).


keep94:
We decided to keep Sender as an interface, so we merged PR #116 instead.

PR #123, opened 2022-11-30
remove unused direct and proxy connection handlers

LukeWinikates:
These two files look to be unused, so it seems helpful to remove them.


LukeWinikates:
@keep94 this is a small change extracted from #120 🙇‍♂️

keep94 (approved):
Good catch finding this dead code. LGTM.

PR #124, opened 2022-11-30
remove an image asset we no longer use

LukeWinikates:
just deleting a small image that used to be displayed in the readme file


keep94 (approved):
LGTM

PR #129, opened 2023-02-16
feat: allow configuring flush interval using time.Duration

LukeWinikates:
see: /issues/121

I think we should keep the `FlushIntervalSeconds` version for now, but probably mark it as deprecated. I agree with @oppegard 's point in the linked issue that semantically it is better to express durations as `time.Duration` rather than as `int`s.

### DIFF
--- a/senders/client.go
+++ b/senders/client.go
@@ -2,15 +2,13 @@ package senders
 
 import (
 	"fmt"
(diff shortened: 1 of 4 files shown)


oppegard (approved):
Looks good – thanks!

Issue #133, opened 2023-04-28
Wavefront URL path is stripped with 0.12.0 sdk version


pavanrangain:
We updated telegraf to version 1.26.2 containing update of Wavefront sdk from 0.11.0 to 0.12.0 and telegraf calls to wavefront proxy are now failing with 404.
We use a Client side load balancer to proxy calls to wavefront proxy instances. Hence our Wavefront URL given to teegraf outputs is as below
http://localhost:29999/wavefront-proxy

With 0.12.0 the path in the URL gets stripped and the proxy calls fail with 404

with 0.11.0 the report URL used to be as below
http://localhost:29999/wavefront-proxy/report?f=wavefront

with 0.12.0 the report URL is as below
http://localhost:29999/report?f=wavefront
(the path parameter gets stripped)

the change in https://github.com/wavefrontHQ/wavefront-sdk-go/pull/122 seem to have caused the issue
esp these lines https://github.com/wavefrontHQ/wavefront-sdk-go/blob/master/senders/client_factory.go#L145-L146 where host and port are considered but the path in base URL gets stripped off

LukeWinikates:
@pavanrangain thanks for taking the time to track down those specific lines. I agree with your assessment - the change in #122 unintentionally removed the path handler.

I'll put together a fix and get it released this week.

LukeWinikates:
@pavanrangain I just opened a pull request with the fix for this issue. Please feel free to have a look and comment if you have any thoughts.

LukeWinikates:
@pavanrangain I think this is resolved now - I'm going to close this issue, but please reopen it if you're still having problems.

PR #134, opened 2023-05-01
fix: restore support for paths in URLs

LukeWinikates:
addresses #133 

This PR restores support for paths in URLs passed to the Wavefront sender.

* adds a top-level integration test for path support, so that it's less likely to break in the future
* makes several helper funcs internal
* moves sender package tests into sender package (so they can call non-public funcs)


jbooherl:
lgtm

PR #141, opened 2023-07-24
chore(ci): user newer github action versions

LukeWinikates:


suprajanarasimhan:
Bumping versions for GitHub actions steps make sense. Looks good!

For my learning: I noticed that there's a v4 version of `actions/setup-go`. However I am guessing that it's better practice to bump one major version (i.e. from v2 to v3) rather than two major versions at once. How do you think about it, @LukeWinikates?

suprajanarasimhan (approved):
Bumping versions for GitHub actions steps make sense. Looks good to me.

For my learning: I noticed that there's a v4 version of actions/setup-go. However I am guessing that it's better practice to bump one major version (i.e. from v2 to v3) rather than two major versions at once. How do you think about it, @LukeWinikates?

LukeWinikates:
Ooh, let's try the new version! I just didn't know about it. I think in this instance, the risk of upgrading is very small. Thanks for taking a close look at the version!

PR #143, opened 2023-07-24
feat: allow users to opt out of internal SDK metrics

LukeWinikates:
See #35


senders/client_factory.go
```diff
@@ -138,16 +147,19 @@ func createConfig(wfURL string, setters ...Option) (*configuration, error) {
...
-// newWavefrontClient creates a Wavefront sender
-func newWavefrontClient(cfg *configuration) (Sender, error) {
+func newSender(cfg *configuration) (Sender, error) {
 	client := internal.NewClient(cfg.Timeout, cfg.TLSConfig)
 	metricsReporter := internal.NewReporter(cfg.metricsURL(), cfg.Token, client)
 	tracesReporter := internal.NewReporter(cfg.tracesURL(), cfg.Token, client)
 	sender := &wavefrontSender{
 		defaultSource: internal.GetHostname("wavefront_direct_sender"),
 		proxy:         !cfg.Direct(),
 	}
```
suprajanarasimhan, line 157: Description: If internal metrics are enabled , create a `RealMetricsRegistry`. Otherwise create a `NoOpRegistry`.

internal/noop_registry.go
```diff
@@ -0,0 +1,85 @@
+package internal
```
suprajanarasimhan, line 1: Introduces the `NoOpRegistry` struct and methods.

internal/real_metric_registry.go
```diff
@@ -0,0 +1,234 @@
+package internal
```
suprajanarasimhan, line 1: This file stores the `RealMetricRegistry` struct and its methods.

internal/registry.go
```diff
@@ -15,130 +10,32 @@ type internalSender interface {
...
-
-func (registry *MetricRegistry) NewGauge(name string, f func() int64) *FunctionalGauge {
-	return registry.getOrAdd(name, &FunctionalGauge{value: f}).(*FunctionalGauge)
+type Incrementer interface {
+	Inc()
 }
 
-func (registry *MetricRegistry) NewGaugeFloat64(name string, f func() float64) *FunctionalGaugeFloat64 {
-	return registry.getOrAdd(name, &FunctionalGaugeFloat64{value: f}).(*FunctionalGaugeFloat64)
-}
```
suprajanarasimhan, line 93: The code below replaces the `MetricRegistry` struct with an interface instead. This is what allows the `configuration` struct field named `registry` to return either (1) a `RealMetricRegistry` type or a (2) `NoOpRegistry` type.

suprajanarasimhan:
If this PR is too big, I can try to spend time to separate the code changes into more chunks. Since it would take some effort to do this, I was thinking of waiting for a reviewer to determine whether they would prefer chunks.

internal/noop_registry.go
```diff
@@ -0,0 +1,85 @@
...
+
+type NoOpRegistry struct{}
```
LukeWinikates, line 3: Do we see any compiler issues if we leave this unexported (i.e. `noOpRegistry`)?

internal/noop_registry.go
LukeWinikates: First impressions:

✅ this is exactly what we talked about
🤔 This interface still 'smells', because 17 methods is a lot.

I think we should break this interface down - the main question is whether doing it now makes this PR more risky or less risky, and more time consuming or less time consuming.

Here's what I'm thinking

Right now we have:

`*Valid()`
`*Invalid()`
`*Dropped()`

for 5 different types, and it's repetitive.

It looks like each line handler accepts the entire `MetricsRegistry` object, but it only calls the 3 Valid/Invalid/Dropped `Incrementer` specific to one type. For example, `SendSpan` only calls `SpansValid`, `SpansInvalid`, and `SpansDropped`, So I think that gives us something to work with to pass each of the handlers a smaller interface that only has methods for its relevant type.

internal/real_metric_registry.go
```diff
@@ -0,0 +1,234 @@
...
+
+import (
+	"sync"
+	"time"
+)
+
+// metric registry for internal metrics
+type RealMetricRegistry struct {
```
LukeWinikates, line 9: I think we can also make this one unexported

internal/real_metric_registry.go
```diff
@@ -0,0 +1,234 @@
...
+	spanLogsValid   *DeltaCounter
+	spanLogsInvalid *DeltaCounter
+	spanLogsDropped *DeltaCounter
+
+	eventsValid   *DeltaCounter
+	eventsInvalid *DeltaCounter
+	eventsDropped *DeltaCounter
+}
+
+func NewMetricRegistry(sender internalSender, setters ...RegistryOption) *RealMetricRegistry {
```
LukeWinikates, line 41: This should return the interface instead

internal/registry.go
LukeWinikates: I'd like to rename this to `InternalMetricRegistry` or `InternalSDKMetricRegistry` and also rename the structs that implement it so that they align with that.

senders/client_factory.go
LukeWinikates: Now that we have `NewSender` calling `newSender`, I wounder if we can unline `newSender` into `NewSender`

senders/integration_test.go
```diff
@@ -46,3 +48,53 @@ func TestTLSEndToEnd(t *testing.T) {
...
+	waitTill1stInternalMetricsCollection := time.Millisecond * 1500
+	internalMetricsInterval := time.Millisecond * 900
+
+	testServer := startTestServer()
+	defer testServer.Close()
+
+	sender, err := NewSender(testServer.URL, InternalMetricsInterval(internalMetricsInterval))
+	require.NoError(t, err)
+	require.NoError(t, sender.SendMetric("my metric", 20, 0, "localhost", nil))
+	time.Sleep(waitTill1stInternalMetricsCollection)
```
LukeWinikates, line 62: Thinking out loud:

We know we don't like having `time.Sleep` in our tests, but I think it's reasonable to do it in an integration test, since we're accepting some inefficiency in exchange for being able to test like a user does.

I think making the intervals fairly short here is a good approach. If the feedback loops are slow, we could try making the intervals even shorter, or we could expose a way of setting the `time.Ticker` itself directly, and I think that would allow us to make it tick imperatively in our tests.

senders/integration_test.go
```diff
@@ -46,3 +48,53 @@ func TestTLSEndToEnd(t *testing.T) {
...
+	require.NoError(t, sender.Flush())
+	metricLines := testServer.MetricLines
+
+	assert.Equal(t, false, HelperInternalMetricFound(metricLines, "points.valid"))
+	assert.Equal(t, 1, len(metricLines))
+	assert.Equal(t, "\"my-metric\" 20 source=\"localhost\"", metricLines[0])
+	assert.Equal(t, "/report?f=wavefront", testServer.LastRequestURL)
+}
+
+func HelperInternalMetricFound(metricLines []string, internalMetricName string) bool {
```
LukeWinikates, line 92: no need to export this, and you might be able to move it to the test server, so you could call it like...

`testServer.HasReceivedMetric("foo.biz.baz")`

LukeWinikates (commented):
@suprajanarasimhan what do you think about moving a few of these files into a sub-package to make the organization more obvious:

`registry`, `real_metric_registry`, and `noop_registry` could all move from `internal` to `internal/sdkmetrics` or something like that.

internal/registry.go
suprajanarasimhan: I chose `InternalSDKMetricRegistry` since I think the word "SDK" may help with understanding the significance of these metrics.

internal/noop_registry.go
suprajanarasimhan: No issues.

senders/integration_test.go
suprajanarasimhan: I think this change makes the code easier to read, thanks!

internal/real_metric_registry.go
suprajanarasimhan: I made the change. I see that the compiler accepts either `*RealMetricRegistry` or the interface type `InternalSDKMetricRegistry`. Can you help me understand why you prefer the interface return type here?

senders/integration_test.go
suprajanarasimhan: Thanks for looking it over! I found it challenging to find the shortest possible values for: `waitTill1stInternalMetricsCollection` and `internalMetricsInterval` to make this test pass predictably. 
- I wanted to leave enough time for the test server to spin up. 
- And I wanted `waitTill1stInternalMetricsCollection` to be long enough to get a first collection but not a second collection.

To reduce the intervals, I think I the next step could be to benchmark how long it takes for the test server to spin up.

suprajanarasimhan: "expose a way of setting the `time.Ticker` itself directly -- Could you explain this a bit more?

suprajanarasimhan:
> @suprajanarasimhan what do you think about moving a few of these files into a sub-package to make the organization more obvious:
> 
> `registry`, `real_metric_registry`, and `noop_registry` could all move from `internal` to `internal/sdkmetrics` or something like that.

That would really highlight its purpose. Do you mind if I try that move? I've never dealt with sub-packages before and I think it would be a good learning experience. (I don't have time to do it today, but I could revisit this tomorrow morning).

senders/client_factory.go
suprajanarasimhan: Can you explain what "unline" means? I tried keyword searching on the web, but I couldn't find a relevant definition.

suprajanarasimhan:
Thanks a lot for the first round of comments, @LukeWinikates! I think it exposed me to some helpful practices and patterns.

internal/noop_registry.go
suprajanarasimhan: Since the `SuccessTracker` interface has been introduced, I can see that `InternalSDKMetricRegistry` only specifies the internal metric families (e.g. points) involved. The fact that each family has `Valid` ,`Invalid` and `Dropped` versions of the metrics is hidden from `InternalSDKMetricRegistry`. For this reason, I'll click "Resolve conversation."

senders/client_factory.go
LukeWinikates: Oh no! That's a typo. I meant "inline", as in eliminating the unexported `newSender` method and putting that code directly in `NewSender`

senders/integration_test.go
LukeWinikates: Sure!

So, right now, the test starts the server, sends some metrics, and waits for metrics flushing to happen in the background. We need to wait for the flushing to happen, but we also want the test to be fast. And like you said, we can only dial down the wait times so much before the tests start to become flaky.

One thought I had was that instead of:

```go
func SetInterval(interval time.Duration) RegistryOption {
	return func(registry *realMetricRegistry) {
		registry.reportTicker = time.NewTicker(interval)
	}
}
```

we could have

```go
func SetReportTicker(ticker *time.Ticker) RegistryOption {
	return func(registry *realMetricRegistry) {
		registry.reportTicker = ticker
	}
}
```

Then, instead of calling:
`SetInterval(1 * time.Second)`, we would have `SetReportTicker(time.newTicker(1 * time.Second))`.

So then maybe we could do something like this

```go
ticker := time.NewTicker(1 * time.Second)
SetReportTicker(&ticker)

// more setup

ticker.C <- time.Now()

// assert

```

The ticker works by sending messages to a channel, so if we hold the ticker, it looks like we can send messages to the channel too, which would allow us to trigger the metrics flushing deterministically without having to wait.

internal/lines.go
```diff
@@ -33,7 +34,7 @@ type LineHandler struct {
 	Format        string
 	flushTicker   *time.Ticker
 
-	internalRegistry *MetricRegistry
+	internalRegistry sdkmetrics.InternalSDKMetricRegistry
```
LukeWinikates, line 37: I guess now, this could be:

`sdkMetricsRegistry sdkMetrics.Registry`

internal/sdkmetrics/real_metric_registry.go
```diff
@@ -0,0 +1,163 @@
+package sdkmetrics
+
+import (
+	"sync"
+	"time"
+)
+
+// metric registry for internal metrics
+type realMetricRegistry struct {
```
LukeWinikates, line 9: now this could be `realRegistry` since `metric` is in the package name

LukeWinikates: we could also call the file `real_registry.go`

senders/client_factory.go
suprajanarasimhan: Cool, that makes sense.  I'd forgotten about the word "inline" -- thanks for the reminder!

Because `newSender` has so many lines in it, I find it pleasant to look at them separately:
- I think of the exported `NewSender` as creating the `configuration` struct. 
- I think of `newSender` as consuming values on the configuration struct to create reporters and line handlers. 

However I also think it would be pretty easy to break it apart again if ever needed, so I think I'll go with the inlining.

LukeWinikates: I agree, `newSender` is a little long. I don't like having two functions whose names are only different by one character, and  I don't like that whatever makes `NewSender` and `newSender` different, the function names don't shed any light on it. If we can make that situation better, I'd be happy to have the two different functions.

senders/integration_test.go
suprajanarasimhan: Thank you for the clear and detailed explanation! 🙏 I'll try this out.

senders/test_server_test.go
```diff
@@ -71,3 +72,13 @@ func decodeMetricLines(request *http.Request) ([]string, error) {
 func (s *testServer) Close() {
 	s.httpServer.Close()
 }
+
+func (s *testServer) hasReceivedInternalMetric(metricLines []string, internalMetricName string) bool {
```
LukeWinikates, line 76: rather than having a `metricsLines` parameter here, you can use `s.MetricLines` inside the method

senders/client_factory.go
suprajanarasimhan: I hadn't thought of these arguments and I agree with them, too. Thanks for explaining. I inlined. 👍

senders/integration_test.go
suprajanarasimhan: I tried doing the refactor. The next step is to modify the test itself. I can pick that up tomorrow.

suprajanarasimhan: When trying to send a message to the ticker's channel, I get an error `Invalid operation: internalMetricsTicker.C <- time.Now() (send to the receive-only type <- chan Time)`. 

I am interpreting this to mean that I can only receive/read messages on this channel (and not send/write messages). 

I wanted to put my thoughts out there for feedback, but I can keep digging.

suprajanarasimhan: Yeah, exactly, darn!

LukeWinikates: So, here's what I came up with, which is probably simpler than any of my previous suggestions.

I added a `Flush()` method to the `sdkmetrics.Registry` type, which for `realRegistry` just calls `report()`. That allows us to send the internal SDK metrics using just a method call.

The fake ticker could be made to work, but it's a lot of dependency injection to be doing in an integration test, so it's probably better not to go down that route.

suprajanarasimhan (approved):

PR #148, opened 2023-08-18
feat: authentication using CSP

LukeWinikates:


internal/reporter.go
```diff
@@ -77,8 +78,9 @@ func (reporter reporter) buildRequest(format string, body []byte) (*http.Request
 
 	req.Header.Set(contentType, octetStream)
 	req.Header.Set(contentEncoding, gzipFormat)
-	if len(reporter.token) > 0 {
```
LukeWinikates, line 80: I'd like to play with this area a bit...

instead of `len` checks, I wonder if we can pull this out into the authorization service and do something like:

`authorizationService.AddHeaders(request)`

internal/token/token_csp.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	return t.AccessToken
+}
+
+func (t *CspServerToServerTokenService) CallCSP() (string, error) {
+	var oauthPath = "/csp/gateway/am/api/auth/authorize"
+	client := &http.Client{}
+
+	req, err := http.NewRequest("POST", t.CSPBaseUrl+oauthPath, strings.NewReader(url.Values{"grant_type": {"client_credentials"}}.Encode()))
```
LukeWinikates, line 46: nit: multi-line this

internal/token/token_csp.go
```diff
@@ -0,0 +1,160 @@
...
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+
+	var cspResponse CSPAuthorizeResponse
+
+	err = json.Unmarshal(body, &cspResponse)
+
+	if !hasDirectIngestScope(cspResponse.Scope) {
+		log.Println("The CSP response did not find any scope matching 'aoa:directDataIngestion' which is required for Wavefront direct ingestion.")
```
LukeWinikates, line 70: I like the added log statements, but I'm conflicted about creating unexpected log entries for customers and in our unit tests

internal/token/token_csp.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+
+		t.AccessToken = cspAccessToken
+		t.tokenReady = true
+	}
+
+	return t.AccessToken
+}
+
+func (t *CspServerToServerTokenService) CallCSP() (string, error) {
```
LukeWinikates, line 42: rename: something intention-revealing like `GetCSPBearerToken`

internal/token/token_csp.go
```diff
@@ -0,0 +1,160 @@
...
+
+	return retVal
+}
+
+func encodeCSPCredentials(CSPClientId string, CSPClientSecret string) string {
+	return "Basic " + base64.StdEncoding.EncodeToString([]byte(CSPClientId+":"+CSPClientSecret))
+}
+
+// NewCSPServerToServerTokenService returns a TokenService instance where it will call CSP with client credentials to return an access token
+func NewCSPServerToServerTokenService(CSPBaseUrl string, CSPClientId string, CSPClientSecret string) TokenService {
```
LukeWinikates, line 153: maybe make a CSP package so we don't have to say CSP so much in type declarations, etc.

senders/client_factory.go
```diff
@@ -31,6 +34,11 @@ type configuration struct {
 	Server string // Wavefront URL of the form https://<INSTANCE>.wavefront.com
 	Token  string // Wavefront API token with direct data ingestion permission
 
+	// CSP Client Credentials should return an access token with direct data ingestion permissions
```
LukeWinikates, line 37: maybe a struct

senders/client_factory.go
```diff
@@ -62,7 +70,7 @@ type configuration struct {
 }
 
 func (c *configuration) Direct() bool {
```
LukeWinikates, line 72: follow-up: I forget how we use this function, but it's starting to look a little painful

senders/client_factory.go
```diff
@@ -85,9 +93,80 @@ func NewSender(wfURL string, setters ...Option) (Sender, error) {
 		return nil, fmt.Errorf("unable to create sender config: %s", err)
 	}
 
+	var tokenService = token.NewNoopTokenService()
```
LukeWinikates, line 96: try: extract a factory function for these lines

senders/client_factory.go
```diff
@@ -85,9 +93,80 @@ func NewSender(wfURL string, setters ...Option) (Sender, error) {
...
+	sender.histoHandler = newLineHandler(metricsReporter, cfg, internal.HistogramFormat, "histograms", sender.internalRegistry)
+	sender.spanHandler = newLineHandler(tracesReporter, cfg, internal.TraceFormat, "spans", sender.internalRegistry)
+	sender.spanLogHandler = newLineHandler(tracesReporter, cfg, internal.SpanLogsFormat, "span_logs", sender.internalRegistry)
+	sender.eventHandler = newLineHandler(metricsReporter, cfg, internal.EventFormat, "events", sender.internalRegistry)
+
+	sender.Start()
+	return sender, nil
+}
+
+func CSPBaseUrl(CSPBaseUrl string) Option {
```
LukeWinikates, line 136: this should live with the other  Option types

senders/client_factory.go
```diff
@@ -85,9 +93,80 @@ func NewSender(wfURL string, setters ...Option) (Sender, error) {
...
+	return func(cfg *configuration) {
+		cfg.CSPBaseUrl = CSPBaseUrl
+	}
+}
+
+// NewCSPSender creates Wavefront client that uses CSP auth
+func NewCSPSender(wfURL string, cspClientId string, cspClientSecret string, setters ...Option) (Sender, error) {
```
LukeWinikates, line 143: Let's find a way to collapse this into just one NewSender function. CSP tokens can be passed as an Option.

senders/client_factory.go
```diff
@@ -132,14 +211,76 @@ func createConfig(wfURL string, setters ...Option) (*configuration, error) {
...
+	}
+	cfg.Server = u.String()
+
+	for _, set := range setters {
+		set(cfg)
+	}
+	return cfg, nil
+}
+
+func createCSPConfig(wfURL string, cspBaseUrl string, cspClientId string, cspClientSecret string, setters ...Option) (*configuration, error) {
```
LukeWinikates, line 247: figure out how to unify this with createConfig

senders/client_test.go
```diff
@@ -22,16 +19,15 @@ const (
...
 var wfPort string
 var proxyPort string
 
 func TestMain(m *testing.M) {
 
 	directServer := httptest.NewServer(directHandler())
 	proxyServer := httptest.NewServer(proxyHandler())
-	wfPort = extractPort(directServer.URL)
-	proxyPort = extractPort(proxyServer.URL)
+	wfPort = test.ExtractPort(directServer.URL)
```
LukeWinikates, line 29: I need to see what's happening here

test/helpers.go
```diff
@@ -0,0 +1,39 @@
+package test
```
LukeWinikates, line 1: change: no new package just for test helpers.

senders/example_newsender_csp_test.go
```diff
@@ -0,0 +1,18 @@
+package senders_test
```
LukeWinikates, line 1: I think tests just go in the `senders` package.

LukeWinikates (commented):
added several comments

internal/token/token_csp.go
LukeWinikates: change: this should probably return an error

internal/token/token_csp.go
```diff
@@ -0,0 +1,161 @@
...
+					t.mutex.Unlock()
+				}
+			}
+		}()
+	} else {
+		t.ticker.Reset(time.Duration(tickerDelay) * time.Second)
+	}
+}
+
+func (t *CspServerToServerTokenService) getTickerDelay(expiresIn int) int {
```
LukeWinikates, line 133: change: method name

LukeWinikates: I would expect a 'get' to take no arguments and return a state value. I want to look at where this is used and see what its intention is. Maybe 'normalize ticker delay?'

internal/token/token_csp.go
```diff
@@ -0,0 +1,175 @@
...
+	"io"
+	"log"
+	"net/http"
+	"net/url"
+	"strings"
+	"sync"
+	"time"
+)
+
+type CspServerToServerTokenService struct {
```
LukeWinikates, line 15: rename: CSPServerToServer

internal/token/token_csp_test.go
```diff
@@ -0,0 +1,106 @@
...
+	assert.False(t, hasDirectIngestScope("no direct ingest scopes"))
+
+	var scopeString = "external/51d98d2c-3ae1-11ee-be56-0242ac120002/*/aoa:directDataIngestion external/51d98d2c-3ae1-11ee-be56-0242ac120002/aoa:directDataIngestion csp:org_member"
+
+	assert.True(t, hasDirectIngestScope(scopeString))
+	assert.True(t, hasDirectIngestScope("some aoa:*"))
+	assert.True(t, hasDirectIngestScope("some aoa/*"))
+}
+
+func directHandler() http.Handler {
```
LukeWinikates, line 76: rename: CSP handler

senders/example_newsender_csp_test.go
LukeWinikates: Looks like I'm wrong

internal/token/token_csp_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"github.com/stretchr/testify/assert"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+)
+
+func TestMultipleCSPRequests(t *testing.T) {
```
LukeWinikates, line 11: I think we want another test for the behavior when:

* CSP is down (don't spam CSP)
* Scope is invalid (also don't spam CSP)

Maybe both of these should switch to a slower polling interval?
If they stop refreshing, they can't send metrics, so the buffer will build up... what happens?

LukeWinikates:
* needs orgID option
* support `ALL PERMISSIONS` scope

priyaselvaganesan (approved):

PR #167, opened 2023-09-27
chore: enable errcheck linter and resolve linter warnings

LukeWinikates:


suprajanarasimhan (approved):
I see that this change enables a linter for identifying and correcting unchecked errors. 
The two types of corrections I see:
- Assert that code in tests doesn't return errors.
- Add underscores to indicate ignored errors. 

Neat!

jbooherl (approved):

PR #169, opened 2023-10-12
refactor: simplify Reporter interface

LukeWinikates:
This refactor simplifies the `Reporter` interface by pushing a conditional from the caller into the `Report` method and eliminating `ReportEvent` from the interface.

* simplifies code at the call site
* makes the interface smaller
* applies [Tell, Don't Ask](https://martinfowler.com/bliki/TellDontAsk.html)

### DIFF
--- a/internal/reporter.go
+++ b/internal/reporter.go
@@ -32,14 +32,18 @@ func (reporter reporter) Report(format string, pointLines string) (*http.Respons
 		return nil, formatError
 	}
 
+	if format == eventFormat {
+		return reporter.reportEvent(pointLines)
+	}
+
 	requestBody, err := linesToGzippedBytes(pointLines)
 	if err != nil {
-		return &http.Response{}, err
+		return nil, err
 	}
 
 	req, err := reporter.buildRequest(format, requestBody)
 	if err != nil {
-		return &http.Response{}, err
+		return nil, err
 	}
 
 	return reporter.execute(req)
@@ -80,7 +84,7 @@ func (reporter reporter) buildRequest(format string, body []byte) (*http.Request
 	return req, nil
 }
 
-func (reporter reporter) ReportEvent(event string) (*http.Response, error) {
+func (reporter reporter) reportEvent(event string) (*http.Response, error) {
 	if event == "" {
 		return nil, formatError
 	}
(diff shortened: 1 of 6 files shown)


jbooherl (approved):
LGTM

internal/reporter.go
```diff
@@ -90,15 +94,26 @@ func (reporter reporter) buildRequest(format string, body []byte) (*http.Request
...
 
-func (reporter reporter) ReportEvent(event string) (*http.Response, error) {
+func (reporter reporter) reportEvent(event string) (*http.Response, error) {
 	if event == "" {
 		return nil, formatError
 	}
 
 	apiURL := reporter.serverURL + eventEndpoint
-	req, err := http.NewRequest("POST", apiURL, strings.NewReader(event))
+	var body io.Reader
```
LukeWinikates, line 103: I think that the way the code is prior to this PR reflects some kind of logical error.

As written, it will add the `Content-Encoding: gzip` header to event requests that aren't gzipped.

I think that the code is trying to make sure that all requests directly posting to a Wavefront host use `gzip`, but forgot to do the encoding step.

Events are odd - the `Direct` code path ingests them as JSON, but the proxy code path expects lines starting with `@Event`.

I think that we might as well `gzip` all of these all the time.

LukeWinikates:
- [ ] verify that the behavior change for the event sender doesn't pose any problems
  - [ ] Wavefront API should be able to handle sending gzipped events 
  - [ ] Wavefront proxy should be able to handle gzipped event lines

LukeWinikates:
Just wrote some additional tests, sending an event directly to Wavefront and sending an event via my local Wavefront proxy.

Neither approach actually supports gzip - the proxy rejects it like this:

Proxy response to gzipped payload

```
2023-10-26 13:39:35,171 INFO  [AbstractReportableEntityHandler:reject] [2878] blocked input: [WF-300 Cannot parse metric: "�rp-K�+Q04��065�47543��LL!<������<�T�R%����[��|���BIb��Rqrj^bQf��BJfQjr�ny�nI~vj� ��8�Cb source="localhost""; remote: 0:0:0:0:0:0:0:1 [2878]; reason: "Invalid value: ��8�Cb"]
```

WF API response to gzipped JSON

```
{"status":{"result":"ERROR","message":"Illegal character ((CTRL-CHAR, code 31)): only regular white space (\\r, \\n, \\t) is allowed between tokens\n at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 2]","code":400}}
```

Without gzip, both work fine - so I will remove the gzip code path for events in this PR.

LukeWinikates:
@jbooherl @suprajanarasimhan made a couple more changes here. Still basically the same change, but I removed the line that sets the `gzip` header, and I added a bunch of tests. I'll probably merge this tomorrow.

PR #171, opened 2023-10-17
feat: add an option to set the http.Client for the Sender

LukeWinikates:
related to https://github.com/influxdata/telegraf/issues/14090

This PR introduces a new `Option`, `HTTPClient`, allowing users to override the `*http.Client` used by the `Sender`.

This allows a variety of customization options, such as proxy settings, idle connection management, and TLS Options. It is intended to integrate well with Telegraf's http config, which can generate an `http.Client` from a standard `toml` configuration (https://github.com/influxdata/telegraf/blob/master/plugins/common/http/config.go)

This PR ~~reimplements the existing `Timeout` and `TLSConfig` options in terms of `configuration.HTTPClient`, and~~ eliminates the internal `NewClient` factory.

~~feedback request: the PR logs a warning if you try to use more than one of `Timeout`, `TLSConfig`, and `HTTPClient` for the same `Sender`, because subsequent calls will clobber the first call. Curious if anyone thinks this should work differently.~~

I had a change of heart about my initial solution to mixing `Timeout`, `TLSConfig`, and `HTTPClient`. I just pushed a second commit with a different solution. This allows `Timeout` and `TLSConfig` to be used together, but prints a warning if either are used after `HTTPClient`. The intended logic now is that if you set an HTTPClient, the `Sender` will just use that, but you can alternatively provide either, both, or neither of `Timeout` and `TLSConfigOptions` as easier shorthand methods for changing those two settings. In that case, we'll make an `http.Client` for you.

### DIFF
--- a/internal/reporter.go
+++ b/internal/reporter.go
@@ -3,11 +3,9 @@ package internal
 import (
 	"bytes"
 	"compress/gzip"
-	"crypto/tls"
 	"io"
 	"net/http"
 	"strings"
-	"time"
 
 	"github.com/wavefronthq/wavefront-sdk-go/internal/auth"
 )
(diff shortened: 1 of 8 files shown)


internal/reporter.go
```diff
@@ -28,14 +26,6 @@ func NewReporter(server string, tokenService auth.Service, client *http.Client)
 	}
 }
 
-func NewClient(timeout time.Duration, tlsConfig *tls.Config) *http.Client {
```
jbooherl, line 31: I almost want to keep some flavor of this and then add a deprecation warning. The new changes make it impossible to keep previous functionality working without changes to client code. Maybe that is ok. Shrug.

jbooherl (approved):
I agree with the changes made. Was going to make very similar comments to how you changed things!

internal/reporter.go
jbooherl: n/a

jessepye (approved):
lgtm!

PR #174, opened 2023-10-23
chore: Sender refactor

LukeWinikates:
The `NewSender` function currently uses a pattern like this:

```
sender.pointHandler = newLineHandler(metricsReporter, cfg, internal.MetricFormat, "points", sender.internalRegistry)
	sender.histoHandler = newLineHandler(metricsReporter, cfg, internal.HistogramFormat, "histograms", sender.internalRegistry)
	sender.spanHandler = newLineHandler(tracesReporter, cfg, internal.TraceFormat, "spans", sender.internalRegistry)
	sender.spanLogHandler = newLineHandler(tracesReporter, cfg, internal.SpanLogsFormat, "span_logs", sender.internalRegistry)
```

In this pattern, the caller must know that you send `points` with the `metricsReporter`, you use the `points` prefix, and you set the `MetricFormat`.

This PR moves this knowledge into dedicated factory functions for each line type, so that the factory knows the information, and the caller only needs to identify the right kind of handler to create - the factory is responsible for knowing how to create valid handlers.

## idea: keep pushing type-specific logic into the `Handler` abstraction

right now the `Sender` has methods that call a particular formatter and then pass its output to the correct handler. For example, `SendMetric` calls `metric.Line` and then tries to send that with the `sender.pointHandler`. It seems as though we could maybe package the line formatter and the line handler together somehow - why have the `Sender` know about so many `handlers` and have to know which one to use which format with, when we could instead have the handler itself know the format transformation it needs to do, in the same way that it knows what prefixes and headers to put on its HTTP requests?

## idea: the `Handler` is actually a `FormatSpecificBatchSender` or something like that.

The `Handler` name seems to hide the fact that this abstraction is where formatted lines get batched together and eventually sent out.

### DIFF
--- a/internal/handler_factory.go
+++ b/internal/handler_factory.go
@@ -1,0 +1,96 @@
+package internal
+
+import (
+	"time"
+
+	"github.com/wavefronthq/wavefront-sdk-go/internal/sdkmetrics"
+)
+
+type HandlerFactory struct {
+	metricsReporter    Reporter
+	tracesReporter     Reporter
+	flushInterval      time.Duration
+	bufferSize         int
+	lineHandlerOptions []LineHandlerOption
+}
+
+func NewHandlerFactory(
+	metricsReporter,
+	tracesReporter Reporter,
+	flushInterval time.Duration,
+	bufferSize int,
+	registry sdkmetrics.Registry) *HandlerFactory {
+	return &HandlerFactory{
+		metricsReporter: metricsReporter,
+		tracesReporter:  tracesReporter,
+		flushInterval:   flushInterval,
+		bufferSize:      bufferSize,
+		lineHandlerOptions: []LineHandlerOption{
+			SetRegistry(registry),
+		},
+	}
+}
+
+func (f *HandlerFactory) NewPointHandler(batchSize int) *RealLineHandler {
+	return NewLineHandler(
+		f.metricsReporter,
+		metricFormat,
+		f.flushInterval,
+		batchSize,
+		f.bufferSize,
+		append(f.lineHandlerOptions,
+			SetHandlerPrefix("points"))...,
+	)
+}
+
+func (f *HandlerFactory) NewHistogramHandler(batchSize int) *RealLineHandler {
+	return NewLineHandler(
+		f.metricsReporter,
+		histogramFormat,
+		f.flushInterval,
+		batchSize,
+		f.bufferSize,
+		append(f.lineHandlerOptions,
+			SetHandlerPrefix("histograms"))...,
+	)
+}
+
+func (f *HandlerFactory) NewSpanHandler(batchSize int) *RealLineHandler {
+	return NewLineHandler(
+		f.tracesReporter,
+		traceFormat,
+		f.flushInterval,
+		batchSize,
+		f.bufferSize,
+		append(f.lineHandlerOptions,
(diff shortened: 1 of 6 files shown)


senders/new_sender.go
```diff
@@ -28,20 +28,20 @@ func NewSender(wfURL string, setters ...Option) (Sender, error) {
 	} else {
 		sender.internalRegistry = sdkmetrics.NewNoOpRegistry()
 	}
-	sender.pointHandler = newLineHandler(metricsReporter, cfg, internal.MetricFormat, "points", sender.internalRegistry)
-	sender.histoHandler = newLineHandler(metricsReporter, cfg, internal.HistogramFormat, "histograms", sender.internalRegistry)
-	sender.spanHandler = newLineHandler(tracesReporter, cfg, internal.TraceFormat, "spans", sender.internalRegistry)
-	sender.spanLogHandler = newLineHandler(tracesReporter, cfg, internal.SpanLogsFormat, "span_logs", sender.internalRegistry)
-	sender.eventHandler = newLineHandler(metricsReporter, cfg, internal.EventFormat, "events", sender.internalRegistry)
 
+	hf := internal.NewHandlerFactory(
```
jbooherl, line 32: Does it make sense to also include cfg.BatchSize here?

internal/handler_factory.go
```diff
@@ -0,0 +1,96 @@
+package internal
+
+import (
+	"time"
+
+	"github.com/wavefronthq/wavefront-sdk-go/internal/sdkmetrics"
+)
+
+type HandlerFactory struct {
```
jbooherl, line 9: Could the formats be moved to this file as I think this is the only place they are used?

jbooherl:
A further enhancement could be to expand the setters used in NewLineHandler for the command line parameters but that could be another pr too. Just feels like it might help with the readability of the handlers

jbooherl (approved):

jbooherl:
I do like the idea trying to move the line formatting into the Factories in a way. Then the handler could have like a .send that would do all the formatting and send to the correct handler. Is this what your thinking?

internal/handler_factory.go
LukeWinikates: I'll take a look at that - I think that they'll probably go away entirely in the next iteration of refactoring. Those values are only read in two places - once if we need to log an error, and once when we check for a special case about Events.

If we can refactor those checks out, I think the constants can go away entirely.

senders/new_sender.go
LukeWinikates: I moved that one around a couple of times... I put it on the factory constructor at first, but then when I realized that Events have a special case for batch size, I liked the idea of expressing that through the API by letting the other handlers accept a batch size, but not that one.

LukeWinikates:
> I do like the idea trying to move the line formatting into the Factories in a way. Then the handler could have like a .send that would do all the formatting and send to the correct handler. Is this what your thinking?

Yep, I think that direction might work. I'm hoping to try incrementally refactoring towards that through a series of small PRs like this one.

LukeWinikates:
@jbooherl thanks for looking at this PR and approving it! Since I'm confident these changes are safe, I'm going to go ahead and merge it and then in the next couple of days I might open another that keeps building on this.

PR #176, opened 2023-11-09
refactor: type-specific 'handlers'

LukeWinikates:
This PR consolidates a repetitive `trySendWith` pattern in `senders.realSender` into a simpler pattern.

This PR assumes that #175 will be merged first

# Old Pattern

```go
func (sender *realSender) SendMetric(name string, value float64, ts int64, source string, tags map[string]string) error {
	line, err := metric.Line(name, value, ts, source, tags, sender.defaultSource)
	return trySendWith(
		line,
		err,
		sender.pointHandler,
		sender.internalRegistry.PointsTracker(),
	)
}
```

# New Pattern
```go
func (sender *realSender) SendMetric(name string, value float64, ts int64, source string, tags map[string]string) error {
	return sender.pointSender.TrySend(metric.Line(name, value, ts, source, tags, sender.defaultSource))
}
```

In the new pattern, the caller of `trySendWith` only needs to tie together two related things (`metric.Line` and `pointSender`), whereas before they needed to chose correctly 3 times (`metric.Line`, `pointHandler`, `PointsTracker`).

The new pattern also mostly allows us to avoid `line, err` as local variables, so that the `Send*` methods can be as short as 1 line.

# Other opportunistic changes

* rename 'LineHandler' interface to 'BatchBuilder' - the main responsibility of this type seems to be collecting lines for a particular point type into batches of the right size and periodically flushing them. Maybe it should actually be called 'BatchBuffer'?
* moves some line-building logic from `SendDelta` into a new `delta.Line` func
* renames `reporter` to `batchingHTTPReporter` (or maybe it should be `HTTPBatchReporter`)

# Issues
- [ ] need to backfill tests for `delta.Line`
- [ ] `wavefront_sender_test.go` has gotten a little weird - it mocks a 2nd-degree dependency instead of a direct one.
- [ ] finalize or revert renames of `LineHandler` and `reporter`
- [ ] maybe `TypedSender` is not a great name

### DIFF
--- a/internal/handler_factory.go
+++ b/internal/handler_factory.go
@@ -6,32 +6,54 @@ import (
 	"github.com/wavefronthq/wavefront-sdk-go/internal/sdkmetrics"
 )
 
-type HandlerFactory struct {
+type SenderFactory struct {
 	metricsReporter    Reporter
 	tracesReporter     Reporter
 	flushInterval      time.Duration
 	bufferSize         int
-	lineHandlerOptions []LineHandlerOption
+	lineHandlerOptions []BatchAccumulatorOption
+	registry           sdkmetrics.Registry
 }
 
-func NewHandlerFactory(
+func NewSenderFactory(
 	metricsReporter,
 	tracesReporter Reporter,
 	flushInterval time.Duration,
 	bufferSize int,
-	registry sdkmetrics.Registry) *HandlerFactory {
-	return &HandlerFactory{
+	registry sdkmetrics.Registry) *SenderFactory {
+	return &SenderFactory{
+		registry:        registry,
 		metricsReporter: metricsReporter,
 		tracesReporter:  tracesReporter,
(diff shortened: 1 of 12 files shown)


internal/handler_factory.go
```diff
@@ -6,32 +6,54 @@ import (
...
 		flushInterval:   flushInterval,
 		bufferSize:      bufferSize,
-		lineHandlerOptions: []LineHandlerOption{
+		lineHandlerOptions: []BatchAccumulatorOption{
 			SetRegistry(registry),
 		},
 	}
 }
 
-func (f *HandlerFactory) NewPointHandler(batchSize int) *RealLineHandler {
```
jbooherl, line 34: nit: could the batchSize be added to the sender factory? Are there plans to make it configurable by sender?

jbooherl (approved):

